PR #3, opened 2019-08-20
Add core pipeline (slot-chain) implementation

louyuting:
### Special notes for reviews


sczyh30:
👍 I'll take a review these days :)

README.md
```diff
@@ -8,3 +8,9 @@
 ## Contributing
 
 Contributions are always welcomed! Please see [CONTRIBUTING](./CONTRIBUTING.md) for detailed guidelines.
+Sentinel Golang
+
+## core package
```
sczyh30: We may put this to the wiki instead of here :)

sczyh30:
Please squash your commits and make the email of your commits match the email of your GitHub account.

core/node.go
```diff
@@ -0,0 +1,27 @@
+package core
+
+type node interface {
+	IncreasePassRequest(count uint64)
+	IncreaseBlockRequest(count uint64)
+	IncreaseErrorRequest(count uint64)
+	IncreaseSuccessRequest(count uint64)
+
+	TotalRequest() uint64
```
sczyh30: For the `Node` interface, we may refer to the counterpart in Sentinel C++: https://github.com/alibaba/sentinel-cpp/blob/master/sentinel-core/statistic/node/node.h. For Java version it might not be elegant enough for the method signature.

core/slot_chain.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+// StatSlot is responsible for counting all custom biz metrics.
+type StatSlot interface {
+	// OnEntryPass function will be invoked when StatPrepareSlots and RuleCheckSlots execute pass
+	// StatSlots will do some statistic logic, such as QPS、log、etc
+	OnEntryPassed(ctx *Context)
+	// OnEntryBlocked function will be invoked when StatPrepareSlots and RuleCheckSlots execute fail
+	// StatSlots will do some statistic logic, such as QPS、log、etc
+	OnEntryBlocked(ctx *Context)
```
sczyh30: Maybe we need an additional parameter to indicate the type of blocking event (like BlockException)?

core/entry.go
```diff
@@ -0,0 +1,64 @@
...
+	// each entry holds a slot chain.
+	// it means this entry will go through the sc
+	sc *SlotChain
+	// caller node
+	originNode node
+	// current resource node
+	currentNode node
+}
+
+type AsyncEntry struct {
```
sczyh30: This might not be necessary?

sczyh30:
We might also need to think of the invocation-chain scope context (carried between different threads). Maybe we could leverage the Golang [Context](https://blog.golang.org/context) to carry the chain-scope information?

core/entry.go
louyuting: Yes, I will remove this

README.md
louyuting: ok

core/node.go
louyuting: Yep.

core/slot_chain.go
louyuting: I have refined here and add  a blocked event as parameter.

sczyh30:
Thanks for contributing!

PR #6, opened 2019-12-11
possible implementation of datasource extension/property

gorexlv:

A simple implementation for datasource extension, which
attempt to get rid of 'Property Converter' for lacking generic in golang.

Just simply opening the subject for discussion.

### Special notes for reviews


sczyh30:
👍 We could discuss here: #7

PR #9, opened 2019-12-11
Issue8 statistic slot and statistic data structure

### LINKED ISSUE #8: [Feature] Statistic data structure, sliding window
## statistic data structure based on sliding window

Type: *feature request*

basic data structure implemented by sliding window is to record basic metrics, including PASS, BLOCKED, SUCCESS, ERROR, RT.

louyuting:
statistic data structure based on sliding window

Fixes #8

### DIFF
--- a/core/base/slot_chain.go
+++ b/core/base/slot_chain.go
@@ -30,19 +30,22 @@ type RuleCheckSlot interface {
 }
 
 // StatSlot is responsible for counting all custom biz metrics.
+// StatSlot would not handle any panic, and pass up all panic to slot chain
 type StatSlot interface {
 	// OnEntryPass function will be invoked when StatPrepareSlots and RuleCheckSlots execute pass
 	// StatSlots will do some statistic logic, such as QPS、log、etc
 	OnEntryPassed(ctx *EntryContext)
-	// OnEntryBlocked function will be invoked when StatPrepareSlots and RuleCheckSlots execute fail
+	// OnEntryBlocked function will be invoked when StatPrepareSlots and RuleCheckSlots fail to execute
+	// It may be inbound flow control or outbound cir
 	// StatSlots will do some statistic logic, such as QPS、log、etc
-	// blockEvent is a enum{RuleBasedCheckBlockedEvent} indicate the block event
+	// blockError introduce the block detail
 	OnEntryBlocked(ctx *EntryContext, blockError *BlockError)
 	// onComplete function will be invoked when chain exits.
+	// The request may be executed successful or blocked or internal error
 	OnCompleted(ctx *EntryContext)
 }
 
-// SlotChain hold all system Slots and customized slot.
+// SlotChain hold all system slots and customized slot.
 // SlotChain support plug-in slots developed by developer.
 type SlotChain struct {
 	statPres   []StatPrepareSlot
@@ -110,13 +113,14 @@ func (sc *SlotChain) AddStatSlotLast(s StatSlot) {
 	sc.stats = append(sc.stats, s)
 }
 
-// The entrance of Slot Chain
+// The entrance of slot chain
+// Return the TokenResult and nil if internal panic.
 func (sc *SlotChain) Entry(ctx *EntryContext) *TokenResult {
 	// This should not happen, unless there are errors existing in Sentinel internal.
-	// defer to handle it
+	// If happened, need to add TokenResult in EntryContext
 	defer func() {
 		if err := recover(); err != nil {
-			logger.Panicf("Unknown panic in SlotChain, err: %+v\n", err)
+			logger.Panicf("Sentinel internal panic in SlotChain, err: %+v", err)
 			return
 		}
 	}()
@@ -152,7 +156,7 @@ func (sc *SlotChain) Entry(ctx *EntryContext) *TokenResult {
 	if len(ss) > 0 {
 		for _, s := range ss {
 			// indicate the result of rule based checking slot.
-			if ruleCheckRet.Status() == ResultStatusPass {
+			if !ruleCheckRet.IsBlocked() {
 				s.OnEntryPassed(ctx)
 			} else {
 				// The block error should not be nil.
--- a/core/base/stat.go
+++ b/core/base/stat.go
@@ -1,14 +1,24 @@
 package base
 
 type TimePredicate func(uint64) bool
-type MetricEvent int32
 
+type MetricEvent int8
+
+// There are five events to record
+// pass + block == Total
 const (
+	// sentinel rules check pass
 	MetricEventPass MetricEvent = iota
+	// sentinel rules check block
 	MetricEventBlock
+
 	MetricEventComplete
+	// Biz error, used for circuit breaker
 	MetricEventError
-	MetricEventRT
+	// request execute rt, unit is millisecond
+	MetricEventRt
+	// hack for the number of event
+	MetricEventTotal
 )
 
 type StatGetter interface {
@@ -25,20 +35,16 @@ type StatNode interface {
 	MetricItemRetriever
 
 	// total  = pass + blocked
+	GetQPS(event MetricEvent) float64
 	TotalQPS() float64
-	PassQPS() float64
-	BlockQPS() float64
-	CompleteQPS() float64
-	ErrorQPS() float64
+
+	AddRequest(event MetricEvent, count uint64)
+	AddRtAndCompleteRequest(rt, count uint64)
 
 	AvgRT() float64
 	MinRT() float64
-	CurrentGoroutineNum() int32
 
-	AddPassRequest(count uint64)
-	AddRtAndCompleteRequest(rt, count uint64)
-	AddBlockRequest(count uint64)
-	AddErrorRequest(count uint64)
+	CurrentGoroutineNum() int32
 	IncreaseGoroutineNum()
 	DecreaseGoroutineNum()
 
--- a/core/stat/base/leap_array.go
+++ b/core/stat/base/leap_array.go
@@ -31,8 +31,8 @@ func (ww *windowWrap) resetTo(startTime uint64) {
 	ww.windowStart = startTime
 }
 
-func (ww *windowWrap) isTimeInWindow(timeMillis uint64, windowLengthInMs uint32) bool {
-	return ww.windowStart <= timeMillis && timeMillis < ww.windowStart+uint64(windowLengthInMs)
+func (ww *windowWrap) isTimeInWindow(now uint64, windowLengthInMs uint32) bool {
+	return ww.windowStart <= now && now < ww.windowStart+uint64(windowLengthInMs)
 }
 
 func calculateStartTime(now uint64, windowLengthInMs uint32) uint64 {
@@ -116,13 +116,13 @@ func (la *leapArray) currentWindow(bg bucketGenerator) (*windowWrap, error) {
 	return la.currentWindowWithTime(util.CurrentTimeMillis(), bg)
 }
 
-func (la *leapArray) currentWindowWithTime(timeMillis uint64, bg bucketGenerator) (*windowWrap, error) {
-	if timeMillis < 0 {
+func (la *leapArray) currentWindowWithTime(now uint64, bg bucketGenerator) (*windowWrap, error) {
+	if now < 0 {
 		return nil, errors.New("Current time is less than 0.")
 	}
 
-	idx := la.calculateTimeIdx(timeMillis)
-	windowStart := calculateStartTime(timeMillis, la.windowLengthInMs)
+	idx := la.calculateTimeIdx(now)
+	windowStart := calculateStartTime(now, la.windowLengthInMs)
 
 	for { //spin to get the current windowWrap
 		old := la.array.get(idx)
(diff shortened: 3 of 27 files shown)


sczyh30:
The CI indicates a data race problem. Could you please resolve it?

sczyh30:
Also please add some brief notes for your design in the PR description or the related issue :)

louyuting:
I am saying and will resolve it later.

louyuting:
@sczyh30  could you please review this PR?

core/base/slot_chain.go
```diff
@@ -152,7 +161,7 @@ func (sc *SlotChain) Entry(ctx *EntryContext) *TokenResult {
 	if len(ss) > 0 {
 		for _, s := range ss {
 			// indicate the result of rule based checking slot.
-			if ruleCheckRet.Status() == ResultStatusPass {
+			if ruleCheckRet.IsPass() {
```
sczyh30: Here we may need a convention: every final status indicates "pass", except the block status (and other special statuses), in case of bugs that yield alien results.

core/stat/base/array_metric.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/sentinel-group/sentinel-golang/core/base"
+	"github.com/sentinel-group/sentinel-golang/util"
+	"math"
+)
+
+// The basic metric struct,
+// ArrayMetric doesn't store any data and is the wrapper of BucketLeapArray to adapt to different internal window
+// ArrayMetric is used for SentinelRules and BucketLeapArray is used for monitor
+// BucketLeapArray is per resource, and ArrayMetric support only read operation.
+type ArrayMetric struct {
```
sczyh30: Maybe `SlidingWindowMetric` is better?

core/stat/base/leap_array.go
```diff
@@ -0,0 +1,203 @@
...
+			if la.updateLock.TryLock() {
+				old = bg.resetWindowTo(old, windowStart)
+				la.updateLock.Unlock()
+				return old, nil
+			} else {
+				runtime.Gosched()
+			}
+		} else if windowStart < old.windowStart {
+			// should not go through here,
+			return nil, errors.New(fmt.Sprintf("Provided time timeMillis=%d is already behind old.windowStart=%d \n", windowStart, old.windowStart))
```
sczyh30: This control branch might be reserved for some special cases (e.g. when occupying "future" buckets).

core/base/stat.go
```diff
@@ -1,14 +1,24 @@
...
+// pass + block == Total
 const (
+	// sentinel rules check pass
 	MetricEventPass MetricEvent = iota
+	// sentinel rules check block
 	MetricEventBlock
+	// complete + error == pass
 	MetricEventComplete
+	// sentinel internal error
 	MetricEventError
```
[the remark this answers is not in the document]
louyuting: ic, it's a mistake comment~
I would update here immediately

core/base/slot_chain.go
```diff
@@ -110,13 +114,18 @@ func (sc *SlotChain) AddStatSlotLast(s StatSlot) {
...
-	// defer to handle it
+	// If happened, need to add TokenResult in EntryContext
 	defer func() {
 		if err := recover(); err != nil {
-			logger.Panicf("Unknown panic in SlotChain, err: %+v\n", err)
+			logger.Panicf("Sentinel internal panic in SlotChain, err: %+v\n", err)
+			if ctx.StatNode == nil {
+				return
+			}
+			ctx.StatNode.AddErrorRequest(uint64(ctx.Input.AcquireCount))
```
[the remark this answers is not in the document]
louyuting: If internal error, from your perspective, how should we do?

core/stat/base/leap_array.go
louyuting: ic, from your statement, it seems that it is not needed to modify here. Is it right?

core/stat/base/array_metric.go
louyuting: Yes.

core/base/slot_chain.go
louyuting: So, if the token result is not blocked, the rest all are pass. Is it rigth?

core/stat/base/leap_array.go
```diff
@@ -0,0 +1,203 @@
...
+
+func calculateStartTime(now uint64, windowLengthInMs uint32) uint64 {
+	return now - (now % uint64(windowLengthInMs))
+}
+
+// atomic windowWrap array to resolve race condition
+// atomicWindowWrapArray can not append or delete element after initializing
+type atomicWindowWrapArray struct {
+	// The base address for real data array
+	base uintptr
```
sczyh30: Caching `uintptr` is dangerous and is highly not recommended according to the document of `unsafe` package:

> Converting a Pointer to a uintptr produces the memory address of the value pointed at, as an integer. The usual use for such a uintptr is to print it.
> Conversion of a uintptr back to Pointer is not valid in general.
> A uintptr is an integer, not a reference. Converting a Pointer to a uintptr creates an integer value with no pointer semantics. Even if a uintptr holds the address of some object, the garbage collector will not update that uintptr's value if the object moves, nor will that uintptr keep the object from being reclaimed.

We'd better use the `unsafe.Pointer` instead.

References:

- https://golang.org/pkg/unsafe/
- https://go101.org/article/unsafe.html

louyuting: That's cool,

> In general, reflect.SliceHeader and reflect.StringHeader should be used only as *reflect.SliceHeader and *reflect.StringHeader pointing at actual slices or strings, never as plain structs. A program should not declare or allocate variables of these struct types.

I also replace the reflect.StringHeader with safe version. Please see below：
```go
type SliceHeader struct {
	Data unsafe.Pointer
	Len  int
	Cap  int
}
```

core/base/slot_chain.go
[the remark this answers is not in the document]
louyuting: Had discussed offline. If internal panic occurs, sentinel will mark this request pass and log detail.

core/stat/base/slidingwindow_metric.go
```diff
@@ -0,0 +1,150 @@
...
+func (m *SlidingWindowMetric) count(event base.MetricEvent, start, end int, value []*windowWrap) int64 {
+	ret := int64(0)
+	l := len(value)
+	for i := start; i < end; i++ {
+		if i >= l {
+			logger.Panicf("The index(%d) is outbound of len(%d).\n", i, l)
+			continue
+		}
+		w := value[i]
+		counter, ok := w.value.(metricBucket)
```
sczyh30: It should be `w.value.(*metricBucket)`?

core/stat/base/slidingwindow_metric.go
```diff
@@ -0,0 +1,150 @@
...
+	allBuckets := m.real.Values(now)
+	minRt := int64(math.MaxInt64)
+	l := len(allBuckets)
+	for i := start; i < end; i++ {
+		if i >= l {
+			logger.Panicf("The index(%d) is outbound of len(%d).\n", i, l)
+			continue
+		}
+		w := allBuckets[i]
+		counter, ok := w.value.(metricBucket)
```
sczyh30: `w.value.(*metricBucket)`?

core/stat/base/slidingwindow_metric.go
```diff
@@ -0,0 +1,150 @@
...
+	return &SlidingWindowMetric{
+		windowLengthInMs: winLengthInMs,
+		sampleCount:      sampleCount,
+		intervalInMs:     intervalInMs,
+		real:             real,
+	}
+}
+
+// Get the [start index, end index) in BucketLeapArray for SlidingWindowMetric in current time.
+func (m *SlidingWindowMetric) getIndexInParent(now uint64) (startIdx, endIdx int) {
```
sczyh30: Any test cases for this method?
Assuming the leap array `a(interval=ia, sampleCount=sa)` and the wrapper SlidingWindowMetric `m(interval=im, sampleCount=sm)` which share the same `bucketLength` value (in fact the bucketLength can be different). The actual end idx should be `a.calculateTimeIdx(currentTime)` and the start idx should be `endIdx - sm`). For example, if sa=20, sm=10, idx in LeapArray for current time is 4, then the actual range of the values should be (4-10, 4], i.e. `[0,4]+[15,19]`.

core/stat/base/slidingwindow_metric.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (m *SlidingWindowMetric) Total() int64 {
+	return m.Pass() + m.Block()
+}
+
+func (m *SlidingWindowMetric) Pass() int64 {
+	now := util.CurrentTimeMillis()
+	start, end := m.getIndexInParent(now)
+	allBuckets := m.real.Values(now)
+	return m.count(base.MetricEventPass, start, end, allBuckets)
```
sczyh30: Using the original index is dangerous (and incorrect) here. The retrieved `allBuckets` might have been filtered (some deprecated buckets are not included), so using the original idx here might cause index out-of-bound error. Maybe providing a time predicate when extracting all data buckets via the `Values` method is better?

core/stat/base/slidingwindow_metric.go
louyuting: had fixed it

core/stat/base/slidingwindow_metric.go
louyuting: had fixed it

core/stat/base/slidingwindow_metric.go
louyuting: Had add unit test.

core/stat/base/slidingwindow_metric.go
louyuting: Had fixed it

core/stat/base/leap_array.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+func (la *leapArray)ValuesWithConditional(timeMillis uint64, predicate base.TimePredicate) []*windowWrap {
+	if timeMillis <= 0 {
+		return make([]*windowWrap, 0)
+	}
+	ret := make([]*windowWrap, 0)
+	for i := 0; i < la.array.length; i++ {
+		ww := la.array.get(i)
+		if ww == nil || la.isWindowDeprecated(timeMillis, ww) || !predicate(timeMillis) {
```
sczyh30: This should be the predicate for the startTime of the bucket, rather than current time? We leverage the bucketStartTime to check whether the bucket satisfy the condition.

core/stat/base/slidingwindow_metric.go
```diff
@@ -0,0 +1,118 @@
...
+func (m *SlidingWindowMetric) Total() int64 {
+	return m.Get(base.MetricEventPass) + m.Get(base.MetricEventBlock)
+}
+
+func (m *SlidingWindowMetric) Get(event base.MetricEvent) int64 {
+	now := util.CurrentTimeMillis()
+	start, end := m.getTimeInterval(now)
+	satisfiedBuckets := m.real.ValuesWithConditional(now, func(u uint64) bool {
+		return now>=start && now<end
+	})
```
sczyh30: Maybe it should be:

```go
satisfiedBuckets := m.real.ValuesWithConditional(now, func(ws uint64) bool {
	return ws>=start && ws<=end
})
```

core/stat/base/slidingwindow_metric.go
```diff
@@ -0,0 +1,118 @@
...
+
+// Get the [start time, end time) of SlidingWindowMetric for now.
+func (m *SlidingWindowMetric) getTimeInterval(now uint64) (start, end uint64) {
+	//
+	timeId := now / uint64(m.windowLengthInMs)
+	idx :=  uint32(timeId) % m.sampleCount
+	bucketStartTime := calculateStartTime(now, m.windowLengthInMs)
+
+	start = bucketStartTime - uint64(idx * m.windowLengthInMs)
+	end = start + uint64(m.intervalInMs)
```
sczyh30: I'm confused by the approach to calculating the condition range. As the latest idx is consistent with current time, maybe the `end` should be yielded first? Here is my code snippet (may not cover all scenarios):

```go
bucketStartTime := calculateStartTime(now, m.real.WindowLengthInMs())

// bucket range (startTime): [start, end]
end = bucketStartTime
start = end - uint64(m.intervalInMs)
```

My test case: inner LeapArray(ia=2s, sa=20), the wrapper Metric(im=2s, sm=10), current timestamp is 1578476668307, then the range of the **inner bucket start time** should be [1578476666300,1578476668300]

core/stat/base/slidingwindow_metric.go
louyuting: If check [start, end], the final interval is (really interval)+1.
So, use the [start, end)

core/stat/base/slidingwindow_metric.go
```diff
@@ -0,0 +1,141 @@
...
+		intervalInMs:     intervalInMs,
+		real:             real,
+	}
+}
+
+// Get the [start time, end time) of SlidingWindowMetric for now.
+func (m *SlidingWindowMetric) getTimeInterval(now uint64) (start, end uint64) {
+	curWindowWrapStartTime := calculateStartTime(now, m.real.WindowLengthInMs())
+	end = curWindowWrapStartTime
+	start = end - uint64(m.intervalInMs) + 1
```
sczyh30: Just a little things to update: `start = end - uint64(m.intervalInMs) + uint64(m.real.WindowLengthInMs())` (i.e. idx + 1)

I'll update by myself later :)

sczyh30:
Thanks for contributing! I'll rearrange some code later.

PR #11, opened 2019-12-13
Add basic logging abstraction

### LINKED ISSUE #10: [Feature] Unified logger
## Issue Description

unified logger to record log

louyuting:
For MVP, I think we could wrap the SDK log and provide the standard log interface. In future, we can upgrade with minimal cost.

Fixes #10


.gitignore
```diff
@@ -3,6 +3,9 @@
 *.iml
 out/
 
+# VS code
+.vscode
```
sczyh30: There's already exclusion for `.vscode` in `.gitignore` :)

louyuting: I had removed that.

sczyh30:
Thanks!

PR #28, opened 2020-01-17
Issue27 refactor statNode and define property model

### LINKED ISSUE #7: [Feature] Data-source and data property listener
## Issue Description

Type: *feature request*

Data-source and data property listener (`SentinelProperty` and `PropertyListener` in Java). Some considerations:

- We might need to support multiple `SentinelProperty` (with different parsers) shared by the single data-source.

### LINKED ISSUE #27: [Enhancement] refactor base statNode
## Issue Description

refactor statNode

louyuting:
enhance statNode.
Define ReadStat and WriteStat to decouple the node operation.

define basic property update model dynamically.

Fixes #27 and also resolves #7 

### Special notes for reviews


core/stat/resource_node.go
```diff
@@ -45,3 +45,7 @@ func (n *ResourceNode) ResourceType() base.ResourceType {
 func (n *ResourceNode) ResourceName() string {
 	return n.resourceName
 }
+
+func (n *ResourceNode) RealBucketLeapArray() *sbase.BucketLeapArray {
```
sczyh30: Maybe this is not necessary? The inner LeapArray should be invisible to users.

louyuting: It is used in circuit breaker.
```go
resNode := stat.GetOrCreateResourceNode(rule.ResourceName())
bla := resNode.RealBucketLeapArray()
```
when we new SlidingWindowMetric, we need pass parameter "*BucketLeapArray".

louyuting: There may be other ways to implement this requirement.

sczyh30: It's okay for now. Some design needs to be refactored later.

sczyh30:
As this PR contains two parts of features, could you please separate it into two commits? One for Sentinel data property and one for enhancement of statistic module.

sczyh30:
Thanks!

PR #30, opened 2020-02-05
Refactor stat node and polish stat pkg

louyuting:
Add sampleCount/interval->SlidingWindowMetric map to ResourceNode.
polish stat pkg, rename windowWrap to bucketWrap

### Special notes for reviews


sczyh30:
Thanks!

PR #31, opened 2020-02-06
auto recover routine to avoid sentinel interrupt main process.

louyuting:
Add util function: WithRecoverGo,  in order to avoid sentinel interrupt main process when internal panic.

None

### Describe how you did it
None

### Describe how to verify it
None

### Special notes for reviews
None


sczyh30:
Thanks!

PR #33, opened 2020-02-07
Add system adaptive rule implementation

### LINKED ISSUE #19: [Feature] System adaptive rule
## Issue Description

Type: *feature request*

Add support for system adaptive rule (adaptive traffic shaping based on metrics like CPU usage, system load and overall QPS/RT.

sczyh30:

Add basic system adaptive rule implementation, which supports various metric types like inbound QPS, RT, concurrency and system load. It's supported to combine adaptive strategy (e.g. BBR) with all supported metric types.

Resolves #19 

### Describe how you did it

- Add system rule definition. Add a `strategy` field to indicate the adaptive strategy to make it flexible.
- Add collector for collecting system metrics (load and CPU usage) using `shirou/gopsutil`.
- Add rule manager and checking slot implementation.

### Describe how to verify it

Run the demo.

### Special notes for reviews

CPU usage mode is not available now, as we need a mechanism to calculate the correct CPU usage with the raw value from `gopsutil`.

Suggestions are needed to improve the unit test.


core/system/sys_stat.go
```diff
@@ -0,0 +1,73 @@
...
+
+	ssStopChan = make(chan struct{})
+)
+
+func init() {
+	currentLoad.Store(notRetrievedValue)
+	currentCpuUsage.Store(notRetrievedValue)
+
+	ticker := time.NewTicker(1 * time.Second)
+	go func() {
```
louyuting: In order to avoid sentinel internal panic interrupt main progress. It may be better to use auto recover goroutine.
Refer to PR: https://github.com/sentinel-group/sentinel-golang/pull/31/files

sczyh30: Good idea. I've polished the code, please check.

PR #54, opened 2020-02-20
Change module to alibaba/sentinel-golang

louyuting:
Change module to alibaba/sentinel-golang

### Special notes for reviews


sczyh30:
Thanks!

PR #56, opened 2020-02-20
Unify configuration

### LINKED ISSUE #42: unify configuration
## Issue Description

Type: *feature request*

Currently, logging config and general config is separated.
Unifying all configurations makes sense

louyuting:
unify configuration

### Describe how you did it
The priority is: env > yaml > default
That is say, yaml config will override default config and env config will override yaml config.

There is only one parameter "config path" for sentinel initialization. If user don't indicate config path, sentinel will use default config.

Unifying the constant to core/constant, including configuration、global constant variable.

FIX #42 
### Special notes for reviews


core/constant/constant.go
```diff
@@ -0,0 +1,37 @@
+package constant
+
+// configuration
```
sczyh30: Actually I don't think we need a universal constant package here. It might be better to put these constants to their respective packages (to make it clear)?

logging/logging.go
```diff
@@ -36,10 +43,15 @@ func GetDefaultLogger() Logger {
 	return defaultLogger
 }
 
+func ResetDefaultLogger(log *log.Logger, namespace string) {
+	defaultLogger.log = log
```
sczyh30: Maybe we need to check nil here?

core/constant/constant.go
louyuting: > Actually I don't think we need a universal constant package here. It might be better to put these constants to their respective packages (to make it clear)?

I agree,  putting all the constants in one package is a bit radical. I will rearrange related code.

sczyh30:
Nice work. Thanks!

PR #57, opened 2020-02-21
Add support for collecting CPU usage for SystemRule

sczyh30:

Add support for collecting CPU usage for SystemRule.

NONE

### Describe how you did it

Calculate the CPU usage:

- u=user+nice, k=system+irq+sirq, total=u+k+idle+steal+iowait
- usage=(udiff+kdiff)/tdiff, or 0 if tdiff=0
- Unify value to [0.0, 1.0]

### Describe how to verify it

Run the test cases and demo.

### Special notes for reviews

NONE

### DIFF
--- a/core/system/sys_stat.go
+++ b/core/system/sys_stat.go
@@ -2,6 +2,7 @@ package system
 
 import (
 	"github.com/alibaba/sentinel-golang/util"
+	"math"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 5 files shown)


core/system/sys_stat.go
```diff
@@ -48,15 +56,49 @@ func retrieveAndUpdateSystemStat() {
...
+		recordCpuUsage(prevCpuStat, curCpuStat)
+		// Cache the latest CPU stat info.
+		prevCpuStat = curCpuStat
 	}
 	if loadStat != nil {
 		currentLoad.Store(loadStat.Load1)
 	}
 }
 
+func recordCpuUsage(prev, curCpuStat cpu.TimesStat) {
```
louyuting: In order to avoid struct copy, it might be better to use Pointer instead of struct here?

sczyh30: Good idea. I've updated the code and added a `stat.system.collectIntervalMs` conf item. Please check.

Issue #55, opened 2020-02-20
Add integration module for dubbo-go


louyuting:
## Issue Description

Type: *feature request*

### Describe what feature you want
Add integration module for [dubbo-go](https://github.com/apache/dubbo-go) client and server.

### Additional context

Discussions are welcomed!

pantianying:
dubbo-go  provides a filter mechanism, which makes it easy to set a custom filter. We just need to implement the following filter interface：
![image](https://user-images.githubusercontent.com/20589945/74914265-1c9b6a00-53fd-11ea-842c-b661a527ef76.png)
并且在config.Load()函数之前调用 ' SetFilter(name string, v func() filter.Filter) '函数，即可将filter实例划到filter库中。

sczyh30:
Resolved via #60.

PR #60, opened 2020-02-23
Add dubbogo adapter

pantianying:

add dubbo-go adapter
https://github.com/alibaba/sentinel-golang/issues/55

### Special notes for reviews


sczyh30:
Could you please sign the CLA here: https://cla-assistant.io/alibaba/sentinel-golang?pullRequest=60

adapter/dubbogo/consumer_filter.go
```diff
@@ -0,0 +1,50 @@
+package dubbogo
```
sczyh30: Maybe the package name could be `dubbo` directly (to make it concise)?

adapter/dubbogo/consumer_filter.go
```diff
@@ -0,0 +1,50 @@
...
+		}
+	} else {
+		// todo : Need to implement asynchronous current limiting
+		//  unlimited flow for the time being
+	}
+	return invoker.Invoke(invocation)
+}
+
+func (d *consumerFilter) OnResponse(result protocol.Result, _ protocol.Invoker, _ protocol.Invocation) protocol.Result {
+	return result
```
sczyh30: Actually, when the invocation completes (whether successful or with error), we need to call `e.Exit()` to mark it as completed, or the response time, complete count and error count will be lost. We may need to carry the `*SentinelEntry` in the RPC context (if possible) and handle it in `OnResponse` callback.

FYI: [onResponse logic in Java version (1.7.1)](https://github.com/alibaba/Sentinel/blob/1.7.1/sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/BaseSentinelDubboFilter.java#L45)

adapter/dubbogo/consumer_filter.go
```diff
@@ -0,0 +1,50 @@
...
+			result.SetResult(nil)
+			result.SetError(b)
+			return result
+		}
+		_, b = sentinel.Entry(methodResourceName,
+			sentinel.WithResourceType(base.ResTypeRPC),
+			sentinel.WithTrafficType(base.Outbound),
+			sentinel.WithArgs(invocation.Attachments()))
+		if b != nil { // blocked
+			result := &protocol.RPCResult{}
```
sczyh30: Can we also provide a fallback registry where developers could register their own fallback logic when blocked? We could provide the default behavior.

Refer: https://github.com/alibaba/Sentinel/tree/1.7.1/sentinel-adapter/sentinel-apache-dubbo-adapter/src/main/java/com/alibaba/csp/sentinel/adapter/dubbo/fallback

pantianying: please check it

adapter/dubbogo/consumer_filter.go
pantianying: please check it

adapter/dubbo/provider_filter.go
```diff
@@ -0,0 +1,44 @@
...
+	_, b = sentinel.Entry(methodResourceName, sentinel.WithResourceType(base.ResTypeRPC), sentinel.WithTrafficType(base.Inbound), sentinel.WithArgs(invocation.Attachments()))
+	if b != nil { // blocked
+		return providerDubboFallback(ctx, invoker, invocation, b)
+	}
+
+	return invoker.Invoke(ctx, invocation)
+}
+
+func (d *providerFilter) OnResponse(ctx context.Context, result protocol.Result, _ protocol.Invoker, _ protocol.Invocation) protocol.Result {
+	if interfaceEntry := ctx.Value(InterfaceEntryKey); interfaceEntry != nil {
```
sczyh30: The convention of exiting multiple Sentinel entries is in LIFO order (e.g. `entry1 -> entry2 -> exit2 -> exit1`).

adapter/dubbo/consumer_filter.go
```diff
@@ -0,0 +1,56 @@
...
+		interfaceEntry, b = sentinel.Entry(interfaceResourceName, sentinel.WithResourceType(base.ResTypeRPC), sentinel.WithTrafficType(base.Outbound))
+		if b != nil { // blocked
+			return consumerDubboFallback(ctx, invoker, invocation, b)
+		}
+		methodEntry, b = sentinel.Entry(methodResourceName, sentinel.WithResourceType(base.ResTypeRPC), sentinel.WithTrafficType(base.Outbound), sentinel.WithArgs(invocation.Attachments()))
+		if b != nil { // blocked
+			return consumerDubboFallback(ctx, invoker, invocation, b)
+		}
+	} else {
+		// TODO : Need to implement asynchronous current limiting
```
sczyh30: What're the differences between async and sync mode?

pantianying: This is mainly to see that in Java code, async entry is used instead of entry to restrict current. At this stage, Dubbo go has not added async field in attachment, so it will not come to async logic

sczyh30: Okay. Sentinel Go has unified `entry` and `asyncEntry`, so we could use `sentinel.Entry(args)` for both sync and async scenarios.

sczyh30:

PR #65, opened 2020-02-25
Add tracer and polish related code

louyuting:
Add tracer

### Special notes for reviews


api/tracer.go
```diff
@@ -0,0 +1,68 @@
...
+		opts.err = err
+	}
+}
+func AddCount(count int) TracerOption {
+	return func(opts *TracerOptions) {
+		opts.count = count
+	}
+}
+
+func Trace(ctx *base.EntryContext, opts ...TracerOption) {
```
sczyh30: Some suggestions:

- The name `Trace` seems too generic. How about `TraceError`?
- Actually it's more common to pass the `Entry` itself instead of the `EntryContext`, so maybe we could provide two functions here: `TraceError(entry, err, opts...)` and `TraceErrorToCtx(ctx, err, opts...)`. Note that the `err` is required (though not used yet).

louyuting: That's great.
I would polish related code.

sczyh30:
Great. Could you please also add simple test cases for `tracer.go`?

louyuting:
> Great. Could you please also add simple test cases for `tracer.go`?

Ok

sczyh30:
Thanks!

Issue #71, opened 2020-02-28
polish the method LoadRules() in rule_manager.go


hellosmallstone:
## Issue Description

Type: *feature request*

The implementation of LoadRules() in rule_manager.go might be not elegant, and the channel ruleChan seems to have no obvious meaning.

### Describe what feature you want

Call onRuleUpdate() to load rules in LoadRules().

hellosmallstone:
@louyuting @sczyh30 
I will resolve the issue.

PR #72, opened 2020-02-28
polish the method LoadRules() in system rule manager.go (#71)

hellosmallstone:
Remove the channel ruleChan which seems to have no obvious meaning.
LoadRules() call onRuleUpdate() to load system rules.
Add ClearRules() to clear all the previous rules.

Signed-off-by: xu wu <wuxu1103@163.com>

The implementation of LoadRules() in rule_manager.go might be not elegant, and the channel ruleChan seems to have no obvious meaning.

Fixes 71

### Describe how you did it

 Remove the channel ruleChan which seems to have no obvious meaning.
 LoadRules() call onRuleUpdate() to load system rules.
 Add ClearRules() to clear all the previous rules.

### Describe how to verify it

go test runs ok.

### Special notes for reviews
logRuleUpdate() seems to have no obvious meaning, therefore I delete it and add some logs in its caller.

### DIFF
--- a/core/system/rule_manager.go
+++ b/core/system/rule_manager.go
@@ -1,11 +1,11 @@
 package system
 
 import (
-	"encoding/json"
-	"github.com/alibaba/sentinel-golang/logging"
-	"github.com/alibaba/sentinel-golang/util"
-	"github.com/pkg/errors"
 	"sync"
+
+	"github.com/alibaba/sentinel-golang/logging"
+
+	"github.com/pkg/errors"
 )
 
 type RuleMap map[MetricType][]*SystemRule
@@ -16,35 +16,9 @@ var (
 
 	ruleMap    = make(RuleMap, 0)
 	ruleMapMux = new(sync.RWMutex)
-
-	ruleChan     = make(chan []*SystemRule, 10)
-	propertyInit sync.Once
 )
 
-func init() {
-	propertyInit.Do(func() {
-		initRuleRecvTask()
-	})
-}
-
-func initRuleRecvTask() {
-	go util.RunWithRecover(func() {
-		for {
-			select {
-			case rules := <-ruleChan:
-				err := onRuleUpdate(rules)
-				if err != nil {
-					logger.Errorf("Failed to update system rules: %+v", err)
-				}
-			}
-		}
-	}, logger)
-}
-func ClearRules() error {
-	_, err := LoadRules(nil)
-	return err
-}
-
+// GetRules return all the rules
 func GetRules() []*SystemRule {
 	ruleMapMux.RLock()
 	defer ruleMapMux.RUnlock()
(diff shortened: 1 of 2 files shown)


core/system/rule_manager.go
```diff
@@ -55,9 +31,28 @@ func GetRules() []*SystemRule {
 }
 
 // LoadRules loads given system rules to the rule manager, while all previous rules will be replaced.
-func LoadRules(rules []*SystemRule) (bool, error) {
-	ruleChan <- rules
-	return true, nil
+func LoadRules(rules []*SystemRule) bool {
```
louyuting: It's better to pass up the error to the upper layer.

core/system/rule_manager.go
```diff
@@ -55,9 +31,28 @@ func GetRules() []*SystemRule {
...
+	if err := onRuleUpdate(rules); err != nil {
+		logger.Errorf("load rules %v failed: %s", rules, err.Error())
+		return false
+	}
+
+	return true
+}
+
+// ClearRules clear all the previous rules
+func ClearRules() bool {
```
louyuting: Could you please provide the use case for function "ClearRules"?

core/system/rule_manager.go
hellosmallstone: In the current case, the meaning of the two return values is repeated. Ehen bool is true, error is always nil. Therefore I delete the error.
If we need pass up the error to the upper layer, we can delete the bool. What do you think of this way?

core/system/rule_manager.go
```diff
@@ -67,22 +62,13 @@ func onRuleUpdate(rules []*SystemRule) error {
 	defer ruleMapMux.Unlock()
 
 	ruleMap = m
-	logRuleUpdate(m)
-
-	return nil
-}
-
-func logRuleUpdate(m RuleMap) {
```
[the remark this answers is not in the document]
hellosmallstone: The function buildRuleMap ignore the invalid rules and get all the valid rules. I think if the input parameter rules has invalid rules, which means that the parameters is problematic, the function should discard all the rules and return error. 
We shouldn't just discard the invalid rule and go against the original will of the caller, even though it is a wrong will.

core/system/rule_manager.go
```diff
@@ -55,9 +31,28 @@ func GetRules() []*SystemRule {
...
+		logger.Errorf("load rules %v failed: %s", rules, err.Error())
+		return false
+	}
+
+	return true
+}
+
+// ClearRules clear all the previous rules
+func ClearRules() bool {
+	if err := onRuleUpdate(nil); err != nil {
```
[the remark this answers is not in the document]
hellosmallstone: I think the operation of ruleMap should be centralized, that is, modify it with one function, and try not to modify it with multiple functions. This seems to have some benefits for others to read the code.
    `func onRuleUpdate(r RuleMap) error` 
We can change the input parameter's type from []*SystemRule to RuleMap, and the funtion will no longer call function buildRuleMap. In this way, the meaning of the function will be more clear.

core/system/rule_manager.go
hellosmallstone: In my design, the LoadRules function will discard empty rules, but the original design can set ruleMap to be empty. Therefore, I provided the ClearRules function to do it specially.
For the use case of the function, I haven't thought much about it yet. We can have some discussion about this.

core/system/rule_manager.go
```diff
@@ -55,9 +31,28 @@ func GetRules() []*SystemRule {
...
-	ruleChan <- rules
-	return true, nil
+func LoadRules(rules []*SystemRule) bool {
+	if len(rules) == 0 {
+		logger.Debug("the rules to be loaded is empty")
+		return true
+	}
+
+	if err := onRuleUpdate(rules); err != nil {
+		logger.Errorf("load rules %v failed: %s", rules, err.Error())
```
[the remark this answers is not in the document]
hellosmallstone: Ok, I will fix it.

core/system/rule_manager.go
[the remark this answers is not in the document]
hellosmallstone: @louyuting Hi, what do you think about my thoughts?

core/system/rule_manager.go
sczyh30: Actually, the `bool` here was designed to indicate whether the internal map has been changed, not for the error.

core/system/rule_manager.go
```diff
@@ -55,9 +31,28 @@ func GetRules() []*SystemRule {
 }
 
 // LoadRules loads given system rules to the rule manager, while all previous rules will be replaced.
-func LoadRules(rules []*SystemRule) (bool, error) {
-	ruleChan <- rules
-	return true, nil
+func LoadRules(rules []*SystemRule) bool {
+	if len(rules) == 0 {
+		logger.Debug("the rules to be loaded is empty")
+		return true
```
sczyh30: It's good to have a `ClearRules` method, but we need to support clear rules by passing an empty slice to `LoadRules` method.

louyuting: It seems that the problem had been fixed.

hellosmallstone:
@louyuting For your question, I added some descriptions.
If one caller want to update four rules, he writes one of them wrong for some reason. Updating only three rules does not seem to be the intention of the caller.
When updating only one rule, if the rule is wrong, we will clear the original rule. This operation seems to require more discussion.

hellosmallstone:
@sczyh30 TBR, thx.

sczyh30:
Could you please resolve the conflicts?

hellosmallstone:
> Could you please resolve the conflicts?

Yeah, I have fixed the conflicts.

sczyh30:
Thanks for contributing!

PR #78, opened 2020-03-03
Change the implementation of LoadRules in flow rule manager.go

wenxuwan:

The implementation of LoadRules() in rule_manager.go might be not elegant, and the channel ruleChan seems to have no obvious meaning.

#71 

### Describe how you did it

Change the function implementation same with flow rule. And remove the unused code in file.

### Describe how to verify it
unit test is ok

### Special notes for reviews

### DIFF
--- a/core/flow/rule_manager.go
+++ b/core/flow/rule_manager.go
@@ -23,8 +23,6 @@ var (
 	tcGenFuncMap = make(map[ControlBehavior]TrafficControllerGenFunc)
 	tcMap        = make(TrafficControllerMap, 0)
 	tcMux        = new(sync.RWMutex)
-
-	propertyInit sync.Once
 )
 
 func init() {
@@ -69,7 +67,7 @@ func onRuleUpdate(rules []*FlowRule) (err error) {
 }
 
 // LoadRules loads the given flow rules to the rule manager, while all previous rules will be replaced.
-func LoadRules(rules []*FlowRule) (error) {
+func LoadRules(rules []*FlowRule) error {
 	// TODO: rethink the design
 	err := onRuleUpdate(rules)
 	return err
--- a/core/flow/rule_manager.go
+++ b/core/flow/rule_manager.go
@@ -3,9 +3,8 @@ package flow
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/pkg/errors"
 	"github.com/alibaba/sentinel-golang/logging"
-	"github.com/alibaba/sentinel-golang/util"
+	"github.com/pkg/errors"
 	"sync"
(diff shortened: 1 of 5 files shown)


core/system/rule_manager.go
```diff
@@ -55,12 +34,22 @@ func GetRules() []*SystemRule {
 }
 
 // LoadRules loads given system rules to the rule manager, while all previous rules will be replaced.
-func LoadRules(rules []*SystemRule) (bool, error) {
```
louyuting: The code related to system rule manager is duplicated  with https://github.com/alibaba/sentinel-golang/pull/72.
I will review flow rule related code.
It's better to remove system rule related code.
Thanks

core/flow/rule_manager.go
```diff
@@ -25,15 +24,10 @@ var (
 	tcMap        = make(TrafficControllerMap, 0)
 	tcMux        = new(sync.RWMutex)
 
-	ruleChan     = make(chan []*FlowRule, 10)
 	propertyInit sync.Once
```
louyuting: You should delete sync.Once simultaneously.

core/flow/rule_manager.go
```diff
@@ -89,10 +67,10 @@ func onRuleUpdate(rules []*FlowRule) (err error) {
 }
 
 // LoadRules loads the given flow rules to the rule manager, while all previous rules will be replaced.
-func LoadRules(rules []*FlowRule) (bool, error) {
+func LoadRules(rules []*FlowRule) error {
```
louyuting: The pair(bool, error) of return value is required.
bool: indicate whether this Load update the rule map;
error: indicate whether occur error in this load operation.

core/system/rule_manager.go
wenxuwan: done

core/flow/rule_manager.go
louyuting: We might add some cache logic to check whether current value is consistent with last update value. It's under discussion.

wenxuwan: But in existing code logic, it is meaningless. We return it directly with True.

wenxuwan: Ok, i recover the code for future design.

louyuting: Currently, we return it directly with True.
As comments upper, it's sure that we will add a cache logic, but the exact design hasn't been decided yet. 
In order to keep all logic consistent in xxx rule manager, it's better to use a pair(bool, error) return value.

louyuting: Thanks~

PR #86, opened 2020-03-10
refreshable file data source

louyuting:
refreshable file data source implementation.

1.  enhance the datasource framework:
    a) Add return value error for PropertyConverter function to indicate whether occur error when converting.
    b)Add return value error for datasource. ReadSource function to indicate whether occur error when reading source.

2. refreshable file datasource implementation

3. add plugin module.
    a) This module is design to encapsulate converter(default json) and updater related logic.

### Describe how to verify it
Unit test

### Special notes for reviews

### DIFF
--- a/ext/datasource/refreshable_file/file_datasource.go
+++ b/ext/datasource/refreshable_file/file_datasource.go
@@ -114,6 +114,6 @@ func (s *RefreshableFileDataSource) doReadAndUpdate() error {
 
 func (s *RefreshableFileDataSource) Close() error {
 	s.closeChan <- struct{}{}
-	logger.Info("The RefreshableFileDataSource had been closed.")
+	logger.Infof("The RefreshableFileDataSource for [%s] had been closed.", s.sourceFilePath)
 	return nil
 }
--- a/ext/datasource/refreshable_file/file_datasource.go
+++ b/ext/datasource/refreshable_file/file_datasource.go
@@ -1,0 +1,119 @@
+package refreshable_file
+
+import (
+	"github.com/alibaba/sentinel-golang/ext/datasource"
+	"github.com/alibaba/sentinel-golang/logging"
+	"github.com/alibaba/sentinel-golang/util"
+	"github.com/fsnotify/fsnotify"
+	"github.com/pkg/errors"
+	"io/ioutil"
+	"os"
+)
+
+var (
+	logger = logging.GetDefaultLogger()
+)
+
+type RefreshableFileDataSource struct {
+	datasource.Base
+	sourceFilePath string
+	isInitialized  util.AtomicBool
+	closeChan      chan struct{}
+	watcher        *fsnotify.Watcher
+}
+
+func NewFileDataSource(sourceFilePath string, handlers ...datasource.PropertyHandler) (*RefreshableFileDataSource, error) {
+	var ds = &RefreshableFileDataSource{
+		sourceFilePath: sourceFilePath,
+		closeChan:      make(chan struct{}),
+	}
+	for _, h := range handlers {
+		ds.AddPropertyHandler(h)
+	}
+	return ds, ds.Initialize()
+}
+
+func (s *RefreshableFileDataSource) ReadSource() ([]byte, error) {
+	f, err := os.Open(s.sourceFilePath)
+	if err != nil {
+		return nil, errors.Errorf("Fail to open the property file, err: %+v.", errors.WithStack(err))
+	}
+	defer f.Close()
+
+	src, err := ioutil.ReadAll(f)
+	if err != nil {
+		return nil, errors.Errorf("Fail to read file, err: %+v.", errors.WithStack(err))
+	}
+	return src, nil
+}
+
+func (s *RefreshableFileDataSource) Initialize() error {
+	if !s.isInitialized.CompareAndSet(false, true) {
+		return nil
+	}
+
+	err := s.doReadAndUpdate()
+	if err != nil {
+		logger.Errorf("Fail to execute doReadAndUpdate, err: %+v", err)
+	}
+
+	w, err := fsnotify.NewWatcher()
+	if err != nil {
+		return errors.Errorf("Fail to new a watcher instance of fsnotify, err: %+v", err)
+	}
+	err = w.Add(s.sourceFilePath)
+	if err != nil {
+		return errors.Errorf("Fail add a watcher on file(%s), err: %+v", s.sourceFilePath, err)
+	}
+	s.watcher = w
+
+	go util.RunWithRecover(func() {
+		defer s.watcher.Close()
+		for {
+			select {
+			case ev := <-s.watcher.Events:
+				if ev.Op&fsnotify.Write == fsnotify.Write {
+					err := s.doReadAndUpdate()
+					if err != nil {
+						logger.Errorf("Fail to execute doReadAndUpdate, err: %+v", err)
+					}
+				}
+
+				if ev.Op&fsnotify.Remove == fsnotify.Remove || ev.Op&fsnotify.Rename == fsnotify.Rename {
+					logger.Errorf("The file source(%s) was removed or renamed.", s.sourceFilePath)
+					for _, h := range s.Handlers() {
+						err := h.Handle(nil)
+						if err != nil {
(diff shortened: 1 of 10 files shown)


ext/datasource/refreshable_file/file_datasource.go
```diff
@@ -0,0 +1,108 @@
...
+	for _, h := range s.Handlers() {
+		err := h.Handle(src)
+		if err != nil {
+			logger.Errorf("RefreshableFileDataSource fail to publish rules, handle: %+v; err: %+v.", h, err)
+		}
+	}
+}
+
+func (s *RefreshableFileDataSource) Close() error {
+	return nil
```
sczyh30: The `Close` op should be supported.

ext/datasource/refreshable_file/file_datasource.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+type RefreshableFileDataSource struct {
+	datasource.Base
+	sourceFilePath string
+	isInitialized  util.AtomicBool
+	watcher        *fsnotify.Watcher
+}
+
+func FileDataSourceStarter(sourceFilePath string, handlers ...datasource.PropertyHandler) (*RefreshableFileDataSource, error) {
```
sczyh30: The name `starter` might be confusing?

ext/datasource/refreshable_file/file_datasource.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	for _, h := range handlers {
+		ds.AddPropertyHandler(h)
+	}
+	return ds, ds.Initialize()
+}
+
+func (s *RefreshableFileDataSource) ReadSource() ([]byte, error) {
+	f, err := os.Open(s.sourceFilePath)
+	defer f.Close()
```
sczyh30: This should be put after the error checking?

ext/datasource/refreshable_file/file_datasource.go
```diff
@@ -0,0 +1,108 @@
...
+
+	if err != nil {
+		return nil, errors.Errorf("The rules file is not existed, err: %+v.", errors.WithStack(err))
```
sczyh30: This may not just indicate non-exist, but also other circumstances like permission denied.

ext/datasource/refreshable_file/file_datasource.go
```diff
@@ -0,0 +1,108 @@
...
+			case err := <-s.watcher.Errors:
+				logger.Errorf("Watch err on file(%s), err: %+v", s.sourceFilePath, err)
+				time.Sleep(time.Second * 3)
+			}
+		}
+	}, logger)
+	return nil
+}
+
+func (s *RefreshableFileDataSource) doUpdate() {
```
sczyh30: How about `doReadAndUpdate`?

ext/datasource/refreshable_file/file_datasource.go
```diff
@@ -0,0 +1,108 @@
...
+	src, err := s.ReadSource()
+	if err != nil {
+		logger.Errorf("Fail to read source, err: %+v", err)
+		return
+	}
+	for _, h := range s.Handlers() {
+		err := h.Handle(src)
+		if err != nil {
+			logger.Errorf("RefreshableFileDataSource fail to publish rules, handle: %+v; err: %+v.", h, err)
```
sczyh30: (1) Here we could use the word "data" instead of `rule`. (2) Printing the handler here might not be helpful. Maybe we need to print out the type?

ext/datasource/refreshable_file/file_datasource.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (s *RefreshableFileDataSource) Initialize() error {
+	if !s.isInitialized.CompareAndSet(false, true) {
+		return errors.New("RefreshableFileDataSource had been Initialized.")
+	}
+
+	s.doUpdate()
+	w, err := fsnotify.NewWatcher()
+	if err != nil {
+		return errors.Errorf("Fail to new a w of fsnotify, err: %+v", err)
```
sczyh30: watcher?

ext/datasource/refreshable_file/file_datasource.go
```diff
@@ -0,0 +1,108 @@
...
+	src, err := ioutil.ReadAll(f)
+	if err != nil {
+		return nil, errors.Errorf("Fail to read file, err: %+v.", errors.WithStack(err))
+	}
+	return src, nil
+}
+
+func (s *RefreshableFileDataSource) Initialize() error {
+	if !s.isInitialized.CompareAndSet(false, true) {
+		return errors.New("RefreshableFileDataSource had been Initialized.")
```
sczyh30: Maybe just returning "nil" is enough?

ext/datasource/refreshable_file/file_datasource.go
louyuting: Make sense, I will refine the statement.

ext/datasource/refreshable_file/file_datasource.go
louyuting: Okay

ext/datasource/refreshable_file/file_datasource.go
louyuting: yes, it's caused by rename, I will rename here.

ext/datasource/refreshable_file/file_datasource.go
louyuting: Make sense.

ext/datasource/refreshable_file/file_datasource.go
louyuting: (1) Right；
(2) Is type the datasource type?

ext/datasource/refreshable_file/file_datasource.go
louyuting: I will rename to NewFileDataSource

ext/datasource/refreshable_file/file_datasource.go
louyuting: Right

ext/datasource/refreshable_file/file_datasource.go
louyuting: Yes, it's caused by mis rename. I will fix here.

ext/datasource/refreshable_file/file_datasource.go
louyuting: Make sense.
The semantic of Close is to close the datasource and stop listen on dynamic datasource.

ext/datasource/refreshable_file/file_datasource.go
```diff
@@ -0,0 +1,114 @@
...
+	watcher        *fsnotify.Watcher
+}
+
+func NewFileDataSource(sourceFilePath string, handlers ...datasource.PropertyHandler) (*RefreshableFileDataSource, error) {
+	var ds = &RefreshableFileDataSource{
+		sourceFilePath: sourceFilePath,
+		closeChan:      make(chan struct{}),
+	}
+	for _, h := range handlers {
+		ds.AddPropertyHandler(h)
```
sczyh30: Do we need to check nil here?

ext/datasource/refreshable_file/file_datasource.go
```diff
@@ -0,0 +1,114 @@
...
+		err := h.Handle(src)
+		if err != nil {
+			logger.Errorf("RefreshableFileDataSource fail to publish property, err: %+v.", err)
+		}
+	}
+}
+
+func (s *RefreshableFileDataSource) Close() error {
+	s.closeChan <- struct{}{}
+	logger.Info("The RefreshableFileDataSource had been closed.")
```
sczyh30: Also print the filePath here to make it clear to identify?

ext/datasource/refreshable_file/file_datasource.go
louyuting: I think it's not necessary, if nil, AddPropertyHandler will not append nil.

louyuting: But we can explicitly check nil in advance here

sczyh30: Okay for me if `AddPropertyHandler` has handled the case.

sczyh30:
Maybe we could also add the case for file-deleted event?

sczyh30:
Nice work. Thanks!

PR #87, opened 2020-03-10
Polish code and comment of config and init package

sczyh30:

Polish code and comment of config and init package.

NONE

### Describe how you did it

- Improve logs and fix tiny problems of config and metric log pkg
- Improve init functions
- Polish comments of config, metric log and init package

### Describe how to verify it

Run the test cases

### Special notes for reviews

NONE

### DIFF
--- a/api/init.go
+++ b/api/init.go
@@ -14,13 +14,11 @@ func InitDefault() error {
 }
 
 // Init loads Sentinel general configuration from the given YAML file
-// and initializes Sentinel. Note that the logging module will be initialized
-// using the configuration from system environment or the default value.
+// and initializes Sentinel.
 func Init(configPath string) error {
 	return initSentinel(configPath)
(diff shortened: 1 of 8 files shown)


api/init.go
```diff
@@ -14,13 +14,11 @@ func InitDefault() error {
 }
 
 // Init loads Sentinel general configuration from the given YAML file
-// and initializes Sentinel. Note that the logging module will be initialized
-// using the configuration from system environment or the default value.
+// and initializes Sentinel.
 func Init(configPath string) error {
```
louyuting: Is panic necessary if an error occurs when the sentinel is initializing?

sczyh30: Maybe we could just return the error and let user decide how to handle it.

louyuting: It's ok.

PR #89, opened 2020-03-11
Enhance datasource framework.

louyuting:
Enhance datasource framework.

### Special notes for reviews

### DIFF
--- a/ext/datasource/datasource.go
+++ b/ext/datasource/datasource.go
@@ -16,9 +16,9 @@ type DataSource interface {
 	ReadSource() ([]byte, error)
 	// Initialize the datasource and load initial rules
 	// start listener to listen on dynamic source
-	// panic if initialize failed;
+	// return error if initialize failed;
 	// once initialized, listener should recover all panic and error.
-	Initialize()
+	Initialize() error
 	// Close the data source.
 	io.Closer
 }
--- a/ext/datasource/datasource.go
+++ b/ext/datasource/datasource.go
@@ -12,9 +12,12 @@ type DataSource interface {
 	// Remove specified property handler in current datasource
 	RemovePropertyHandler(h PropertyHandler)
 	// Read original data from the data source.
-	ReadSource() []byte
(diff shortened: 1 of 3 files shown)


ext/datasource/datasource.go
```diff
@@ -12,9 +12,12 @@ type DataSource interface {
...
-	ReadSource() []byte
-	// Initialize, init datasource and load initial rules
-	// start listener
+	// return source bytes if succeed to read, if not, return error when reading
+	ReadSource() ([]byte, error)
+	// Initialize the datasource and load initial rules
+	// start listener to listen on dynamic source
+	// panic if initialize failed;
+	// once initialized, listener should recover all panic and error.
 	Initialize()
```
sczyh30: Do we need to return the error instead of just panicking?

ext/datasource/property.go
```diff
@@ -51,15 +55,18 @@ func (h *DefaultPropertyHandler) Handle(src []byte) error {
 		}
 	}()
 	// converter to target property
-	realProperty := h.converter(src)
+	realProperty, err := h.converter(src)
+	if err!=nil {
```
sczyh30: Plz format the code :)

ext/datasource/property.go
```diff
@@ -51,15 +55,18 @@ func (h *DefaultPropertyHandler) Handle(src []byte) error {
 		}
 	}()
 	// converter to target property
```
sczyh30: Convert to?

ext/datasource/datasource.go
louyuting: Initialize function should be executed automatically when new a DataSource and the user doesn't need to call it manually.

ext/datasource/property.go
louyuting: Yes, I will polish here.

ext/datasource/property.go
louyuting: ok~

sczyh30:
Thanks!

PR #90, opened 2020-03-11
Add helper for system.SystemRule and flow.FlowRule in datasource

louyuting:
Encapsulate converter(default json) and updater related logic for specific property.
properties:

1. []system.SystemRule
2. []flow.FlowRule

### Special notes for reviews

### DIFF
--- a/core/flow/rule_manager.go
+++ b/core/flow/rule_manager.go
@@ -81,12 +81,8 @@ func GetRules() []*FlowRule {
 }
 
 func ClearRules() error {
-	isUpdated, err := LoadRules(nil)
-	if isUpdated {
-		return err
-	} else {
-		return errors.Wrap(err, "Fail to clear rules.")
-	}
+	_, err := LoadRules(nil)
+	return err
 }
 
 func rulesFrom(m TrafficControllerMap) []*FlowRule {
--- a/core/flow/rule_manager.go
+++ b/core/flow/rule_manager.go
@@ -80,6 +80,15 @@ func GetRules() []*FlowRule {
 	return rulesFrom(tcMap)
 }
 
+func ClearRules() error {
+	isUpdated, err := LoadRules(nil)
+	if isUpdated {
+		return err
+	} else {
+		return errors.Wrap(err, "Fail to clear rules.")
+	}
+}
+
 func rulesFrom(m TrafficControllerMap) []*FlowRule {
 	rules := make([]*FlowRule, 0)
 	if len(m) == 0 {
--- a/core/system/rule_manager.go
+++ b/core/system/rule_manager.go
@@ -40,6 +40,14 @@ func initRuleRecvTask() {
 		}
 	}, logger)
 }
+func ClearRules() error {
+	isUpdated, err := LoadRules(nil)
+	if isUpdated {
+		return err
+	} else {
+		return errors.Wrap(err, "Fail to clear rules.")
+	}
+}
 
 func GetRules() []*SystemRule {
 	ruleMapMux.RLock()
@@ -54,8 +62,8 @@ func GetRules() []*SystemRule {
 
 // LoadRules loads given system rules to the rule manager, while all previous rules will be replaced.
 func LoadRules(rules []*SystemRule) (bool, error) {
-	ruleChan <- rules
-	return true, nil
+	err := onRuleUpdate(rules)
+	return true, err
 }
 
 func onRuleUpdate(rules []*SystemRule) error {
--- a/core/system/rule_manager.go
+++ b/core/system/rule_manager.go
@@ -41,12 +41,8 @@ func initRuleRecvTask() {
 	}, logger)
 }
 func ClearRules() error {
-	isUpdated, err := LoadRules(nil)
-	if isUpdated {
-		return err
-	} else {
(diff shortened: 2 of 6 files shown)


ext/datasource/plugin/flow_rule.go
```diff
@@ -0,0 +1,47 @@
...
+	rules := make([]*flow.FlowRule, 0)
+	if data == nil {
+		rules = nil
+	} else {
+		val, ok := data.([]flow.FlowRule)
+		if !ok {
+			return errors.New(fmt.Sprintf("Fail to type assert data to []flow.FlowRule, in fact, data: %+v", data))
+		}
+		for _, v := range val {
+			rules = append(rules, &v)
```
wenxuwan: why not use val directly ?

louyuting: Avoid redundant assignment copies.

ext/datasource/plugin/flow_rule.go
```diff
@@ -0,0 +1,47 @@
...
+	return ret, nil
+}
+
+// FlowRulesUpdater load the newest []flow.FlowRule to downstream flow component.
+func FlowRulesUpdater(data interface{}) error {
+	rules := make([]*flow.FlowRule, 0)
+	if data == nil {
+		rules = nil
+	} else {
+		val, ok := data.([]flow.FlowRule)
```
sczyh30: Maybe here we could accept both `[]flow.FlowRule` and `[]*flow.FlowRule` type?

ext/datasource/plugin/flow_rule.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/alibaba/sentinel-golang/ext/datasource"
+	"github.com/pkg/errors"
+)
+
+// FlowRulesJsonConverter provide JSON  as the default serialization for flow.FlowRule
+func FlowRulesJsonConverter(src []byte) (interface{}, error) {
+	if len(src) == 0 {
+		return nil, nil
+	}
+	ret := make([]flow.FlowRule, 0)
```
sczyh30: Maybe we could use `[]*flow.FlowRule` here?

ext/datasource/plugin/flow_rule.go
```diff
@@ -0,0 +1,47 @@
...
+		val, ok := data.([]flow.FlowRule)
+		if !ok {
+			return errors.New(fmt.Sprintf("Fail to type assert data to []flow.FlowRule, in fact, data: %+v", data))
+		}
+		for _, v := range val {
+			rules = append(rules, &v)
+		}
+	}
+	succ, err := flow.LoadRules(rules)
+	if succ {
```
sczyh30: The `bool` returned by `LoadRules()` does not indicate error, so maybe we could just return `err` here?

core/flow/rule_manager.go
```diff
@@ -80,6 +80,13 @@ func GetRules() []*FlowRule {
 	return rulesFrom(tcMap)
 }
 
+func ResetTcMap() {
```
sczyh30: Use `LoadRules(nil)` instead?

sczyh30: We may not need such `reset` functions actually :)
If necessary, it should be `ClearRules()`.

louyuting: ClearRules() seem make more sense.
What you see is what you get.

ext/datasource/plugin/flow_rule.go
louyuting: make sense

ext/datasource/plugin/flow_rule.go
louyuting: Make sense

core/flow/rule_manager.go
```diff
@@ -80,6 +80,15 @@ func GetRules() []*FlowRule {
 	return rulesFrom(tcMap)
 }
 
+func ClearRules() error {
+	isUpdated, err := LoadRules(nil)
+	if isUpdated {
+		return err
+	} else {
+		return errors.Wrap(err, "Fail to clear rules.")
```
sczyh30: This does not indicate failure, as the rules may have been cleared before. Maybe here we could just return `err`?

core/system/rule_manager.go
```diff
@@ -40,6 +40,14 @@ func initRuleRecvTask() {
 		}
 	}, logger)
 }
+func ClearRules() error {
+	isUpdated, err := LoadRules(nil)
+	if isUpdated {
+		return err
+	} else {
+		return errors.Wrap(err, "Fail to clear rules.")
```
sczyh30: Ditto.

go.mod
```diff
@@ -14,4 +14,5 @@ require (
 	golang.org/x/sys v0.0.0-20200107162124-548cf772de50 // indirect
 	gopkg.in/yaml.v2 v2.2.2
 	go.uber.org/multierr v1.5.0
+	github.com/fsnotify/fsnotify v1.4.7
```
sczyh30: Is it necessary?

louyuting: I will remove it.
Thanks

sczyh30:
Thanks!

PR #92, opened 2020-03-15
Hotfix: Fix bug for SystemRulesUpdater and FlowRulesUpdater

louyuting:
Hotfix: If the update value is nil in PropertyUpdater, sentinel should clear downstream property.


ext/datasource/helper.go
```diff
@@ -24,7 +24,7 @@ func FlowRulesJsonConverter(src []byte) (interface{}, error) {
 // FlowRulesUpdater load the newest []flow.FlowRule to downstream flow component.
 func FlowRulesUpdater(data interface{}) error {
 	if data == nil {
-		return nil
+		return flow.ClearRules()
```
wenxuwan: The ut of helper should be changed.

sczyh30:
Thanks!

PR #95, opened 2020-03-18
Add adapter for echo framework

zhangmingke:
add adapter for echo framework

- SentinelMiddleware returns new echo.HandlerFunc

- Default resource name is {method}:{path}, such as "GET:/api/:id"

-  Default block fallback is returning 429 code

- Define your own behavior by setting options

### Special notes for reviews


sczyh30:
Could you please add some description about your design according to the PR template?

adapter/echo/middleware.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/labstack/echo/v4"
+)
+
+// SentinelMiddleware
+func SentinelMiddleware(opts ...Option) echo.MiddlewareFunc {
+
+	options := evaluateOptions(opts)
+	return func(next echo.HandlerFunc) echo.HandlerFunc {
+		return func(c echo.Context) (err error) {
+			resourceName := c.Request().Method + ":" + c.Request().URL.Path
```
sczyh30: Can we extract the matching API pattern (e.g. `/foo/:id`) instead of the actual URL?

adapter/echo/option.go
```diff
@@ -0,0 +1,42 @@
...
+	optCopy := &options{}
+	for _, opt := range opts {
+		opt(optCopy)
+	}
+
+	return optCopy
+}
+
+// WithResourceExtractor set resourceExtract
+func With(handlerFunc echo.HandlerFunc) Option {
```
sczyh30: Remove the unused code?

adapter/echo/middleware.go
```diff
@@ -0,0 +1,41 @@
...
+				resourceName,
+				sentinel.WithResourceType(base.ResTypeWeb),
+				sentinel.WithTrafficType(base.Inbound),
+			)
+
+			if errEntry != nil {
+				if options.blockFallback != nil {
+					err = options.blockFallback(c)
+				} else {
+					err = c.JSON(http.StatusTooManyRequests, "error")
```
sczyh30: Maybe the default error string could be "Blocked by Sentinel"?

sczyh30:
And could you please resolve conflicts?

adapter/echo/middleware.go
zhangmingke: ok, i have deal with it

sczyh30:

PR #101, opened 2020-03-19
Refine file datasource code and add ut

louyuting:
Refine file datasource code and add ut

### Special notes for reviews


sczyh30:
Thanks!

PR #109, opened 2020-03-23
Polish code in some details

chenjiandongx:

Polish codes in some details, for example

* remove unreachable code
* init a map elegantly, [blog.golang.org/maps](https://blog.golang.org/maps)

### Special notes for reviews


chenjiandongx:
@zsy619 yet another pr.

PR #115, opened 2020-03-27
Etcdv3 datasource

louyuting:

extend etcdv3 as datasource.
1.  Introduce DatasourceGenerator to generate Etcdv3DataSource.
2. All Etcdv3DataSources generated by DatasourceGenerator instance will reuse the etcdv3 client long lived http connection.
3. Etcdv3DataSource could be closed,  if all Etcdv3DataSources generated by DatasourceGenerator instance were clsoed, DatasourceGenerator will be closed, which means etcdv3 client will be closed.
4. DatasourceGenerator only expose the function Generate.

### Special notes for reviews

### DIFF
--- a/ext/datasource/etcdv3/etcdv3.go
+++ b/ext/datasource/etcdv3/etcdv3.go
@@ -2,7 +2,6 @@ package etcdv3
 
 import (
 	"context"
-	"fmt"
 	"github.com/alibaba/sentinel-golang/ext/datasource"
 	"github.com/alibaba/sentinel-golang/logging"
 	"github.com/alibaba/sentinel-golang/util"
@@ -57,10 +56,10 @@ func (s *Etcdv3DataSource) ReadSource() ([]byte, error) {
 	defer cancel()
 	resp, err := s.client.Get(ctx, s.propertyKey)
 	if err != nil {
-		return nil, datasource.NewDSError(datasource.EtcdGetValueError, fmt.Sprintf("%+v", err))
+		return nil, errors.Errorf("Fail to get value for property key[%s]", s.propertyKey)
 	}
 	if resp.Count == 0 {
-		return nil, datasource.NewDSError(datasource.EtcdKeyNotExistedError, fmt.Sprintf("The key[%s] is not existed in etcd server.", s.propertyKey))
+		return nil, errors.Errorf("The key[%s] is not existed in etcd server.", s.propertyKey)
 	}
 	s.lastUpdatedRevision = resp.Header.GetRevision()
 	logger.Infof("Get the newest data for key:%s, revision: %d, value: %s", s.propertyKey, resp.Header.GetRevision(), resp.Kvs[0].Value)
--- a/ext/datasource/etcdv3/etcdv3.go
+++ b/ext/datasource/etcdv3/etcdv3.go
@@ -1,0 +1,137 @@
+package etcdv3
+
+import (
+	"context"
+	"fmt"
+	"github.com/alibaba/sentinel-golang/ext/datasource"
+	"github.com/alibaba/sentinel-golang/logging"
+	"github.com/alibaba/sentinel-golang/util"
+	"github.com/coreos/etcd/clientv3"
+	"github.com/coreos/etcd/mvcc/mvccpb"
+	"github.com/pkg/errors"
+	"time"
+)
+
+var (
+	logger = logging.GetDefaultLogger()
+)
+
+type Etcdv3DataSource struct {
+	datasource.Base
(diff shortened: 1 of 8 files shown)


louyuting:
@sczyh30  Could you please review this PR?
I will add unit test and example after the design was finalized.

gorexlv:
I suppose the EtcdDatasourceGenerator is a bit over-designed.  Consider to resue etcd client instance through enrich configuration, thus removing the Generator.

How about https://github.com/alibaba/sentinel-golang/pull/116/files#diff-d77ed18801d12e79f4e8c13153b63c94 ?

ext/datasource/etcdv3/etcdv3.go
```diff
@@ -0,0 +1,140 @@
...
+			}
+		}
+	}
+}
+
+func (s *Etcdv3DataSource) watch() {
+	// Add watch for propertyKey from lastUpdatedRevision updated after Initializing
+	ctx, cancel := context.WithCancel(context.Background())
+	s.cancel = cancel
+	rch := s.client.Watch(ctx, s.propertyKey, clientv3.WithCreatedNotify(), clientv3.WithRev(s.lastUpdatedRevision))
```
gorexlv: You'd better create new watcher by clientv3.NewWatcher(client) instead using s.client.Watcher

louyuting: In fact, one WatchClient  could watch on multi key range.
Could you please introduce some more details to describe why to create the new watcher for each propertyKey?

louyuting:
> I suppose the EtcdDatasourceGenerator is a bit over-designed. Consider to resue etcd client instance through enrich configuration, thus removing the Generator.
> 
> How about https://github.com/alibaba/sentinel-golang/pull/116/files#diff-d77ed18801d12e79f4e8c13153b63c94 ?

The idea is pretty good. I will rethink about generator.

ext/datasource/etcdv3/etcdv3.go
```diff
@@ -0,0 +1,150 @@
...
+		s.lastUpdatedRevision = resp.Header.GetRevision()
+	}
+
+	if err := resp.Err(); err != nil {
+		logger.Errorf("Watch on etcd endpoints(%+v) occur error, err: %+v", s.client.Endpoints(), err)
+		return
+	}
+
+	for _, ev := range resp.Events {
+		fmt.Printf("%s %q : %q\n", ev.Type, ev.Kv.Key, ev.Kv.Value)
```
louyuting: It's for example test, I will remove here later.

ext/datasource/etcdv3/etcdv3.go
```diff
@@ -0,0 +1,150 @@
...
+		ds.AddPropertyHandler(h)
+	}
+	return ds, nil
+
+}
+
+func (s *Etcdv3DataSource) ReadSource() ([]byte, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	resp, err := s.client.Get(ctx, s.propertyKey)
```
louyuting, line 49: In order to troubleshooting, we should add a info log to log resp

ext/datasource/etcdv3/etcdv3.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if resp.Count == 0 {
+		return nil, errors.Errorf("The key[%s] is not existed in etcd.", s.propertyKey)
```
louyuting, line 54: Need to return (nil, nil)

ext/datasource/etcdv3/etcdv3.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	s.lastUpdatedRevision = resp.Header.GetRevision()
+	return resp.Kvs[0].Value, nil
+}
+
+func (s *Etcdv3DataSource) Initialize() error {
+	err := s.doReadAndUpdate()
```
louyuting, line 61: Should not return error if parser occurs error when initializing

ext/datasource/etcdv3/etcdv3.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return err
```
sczyh30, line 63: Here we might need to properly handle the error (e.g. parse error) in case of getting in the way of initialization.

sczyh30:

louyuting:
Thanks for @wenxuwan 
Your idea is very helpful.

Issue #108, opened 2020-03-23
[BUG]add len(items) > 0 for getLatestSecond function


zsy619:
## Issue Description

Type: *bug report*

### Describe what happened
add test method happend：
panic: runtime error: index out of range [-1] [recovered]
	panic: runtime error: index out of range [-1]

### Describe what you expected to happen
test code：
make param items equals []*base.MetricItem{}

### Anything else we need to know?
modify ：
if items == nil {
		return 0
	}
to：
if items == nil  || len(items) == 0{
		return 0
	}

louyuting:
Fixed by https://github.com/alibaba/sentinel-golang/pull/121

PR #121, opened 2020-04-08
[BUG]add len(items) > 0 for getLatestSecond function

zsy619:
#108 
### Special notes for reviews


core/log/metric/reader.go
```diff
@@ -174,7 +175,7 @@ func readLine(bufReader *bufio.Reader) (string, error) {
 }
 
 func getLatestSecond(items []*base.MetricItem) uint64 {
-	if items == nil {
+	if items == nil || len(items) == 0 {
```
louyuting, line 178: len(items) == 0  is OK.
The length of nil slice is 0.

hellosmallstone: Yes, the length of nil slice is 0. 
However, the zero value for slice is nil. It seems necessary to standardize the way to return an empty slice in the project. Even in the same function, there are different ways to return empty slice, such as function ReadMetricsByEndTime in this file.

louyuting: Yes, your idea is great.
Could you please d push a PR to enhance.

louyuting:
@zsy619  Any update?

PR #124, opened 2020-04-10
make EntryOptions support storing custom data before Sentinel.Entry()

novoland:

We need to pass some data about current invocation to slots, while these data is parsed before `Sentinel.Entry()`. `EntryOptions` is a suitable place but extra attribute is required.

None.

### Describe how you did it

Add a map `data map[string]interface{}` (and several functions associated) to `EntryOptions`.

### Special notes for reviews

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -12,6 +12,7 @@ type EntryOptions struct {
 	flag         int32
 	slotChain    *base.SlotChain
 	args         []interface{}
+	data         map[string]interface{}
 }
 
 type EntryOption func(*EntryOptions)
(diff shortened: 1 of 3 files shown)


api/api.go
```diff
@@ -12,6 +12,7 @@ type EntryOptions struct {
 	flag         int32
 	slotChain    *base.SlotChain
 	args         []interface{}
+	attachment   map[interface{}]interface{}
```
sczyh30, line 15: Maybe `attachments` is better?

core/base/context.go
```diff
@@ -36,7 +36,7 @@ type SentinelInput struct {
 	Args         []interface{}
 
 	// store some values in this context when calling context in slot.
-	data map[interface{}]interface{}
+	Attachment map[interface{}]interface{}
```
sczyh30, line 39: Ditto

sczyh30:
Thanks for contributing!

PR #138, opened 2020-04-26
add max concurrency record

ansiz:

add max concurrency record, sometimes we need a max concurrency count instead of current instantaneous concurrency count

### Describe how you did it

add max concurrency record, sometimes we need a max concurrency count instead of current instantaneous concurrency count

### Special notes for reviews


sczyh30:
Thanks!

Issue #137, opened 2020-04-24
NewTokeResult exist a performance problem，recommended sync.pool


NineSunRD:
![image](https://user-images.githubusercontent.com/13688621/80175867-743a9980-8629-11ea-9c4e-6e808d977b0c.png)

louyuting:
Hi @NineSunRD 
Your advice is very helpful. Could you please contribute for this issue?
Waiting for your amazing PR.
Thanks.

NineSunRD:
It is my pleasure to contribute for this issue.
I will use my spare time to contribute for this issue.
Thanks.

louyuting:
Done, refer to PR: https://github.com/alibaba/sentinel-golang/pull/142

PR #142, opened 2020-04-28
Add result token pool  (#137)

NineSunRD:
I want to modify the original code as little as possible, but now it exists a lot of temporary token result, so my modify will look very ugly, as shown below。
image

#137 

### Special notes for reviews

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -87,6 +87,13 @@ func Entry(resource string, opts ...EntryOption) (*base.SentinelEntry, *base.Blo
 }
 
 func entry(resource string, options *EntryOptions) (*base.SentinelEntry, *base.BlockError) {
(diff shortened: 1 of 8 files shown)


louyuting:
The CI build failed. Could you please fix it?

sczyh30:
Thanks for contributing! However this seems not elegant enough and may bring boilerplate code. We need to open a new issue to improve this.

sczyh30:
And looking forward to more contributions :)

louyuting:
> Thanks for contributing! However this seems not elegant enough and may bring boilerplate code. We need to open a new issue to improve this.

Yes, It should be improve.

NineSunRD:
Yes, I think so, I have seen the new issue, I will try to contribute it。

PR #143, opened 2020-04-28
Circuit breaker refactor and enhance the scalability of stat module

louyuting:
### Describe what this PR does / why we need it
Circuit breaker refactor and enhance the scalability of stat module

### Special notes for reviews

PR #153, opened 2020-05-22
Polish api.Tracer with new error tracing mechanism

sczyh30:

Polish `api.Tracer` with new error tracing mechanism (set error to entry and record on request completed).

Related to #143

### DIFF
--- a/api/tracer.go
+++ b/api/tracer.go
@@ -2,86 +2,22 @@ package api
 
 import (
 	"github.com/alibaba/sentinel-golang/core/base"
-	"github.com/alibaba/sentinel-golang/core/stat"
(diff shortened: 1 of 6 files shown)


api/tracer.go
```diff
@@ -2,86 +2,22 @@ package api
...
-	}
-	for _, opt := range opts {
-		opt(&options)
-	}
-
-	traceErrorToNode(node, err, options.count)
+	entry.SetError(err)
 }
 
 func traceErrorToNode(node base.StatNode, err error, cnt uint64) {
```
louyuting, line 23: Is there anything else I need to use this function for？If not, we might need delete this func.

sczyh30: Removed/

Issue #161, opened 2020-06-18
support setting global config directly


gorexlv:
## Issue Description

Type: *feature request*

### Describe what feature you want

Currently, sentinel-golang support setting global config by passing configFilePath(must be yaml, btw) or setting environments, which is not friendly to web framework or infrastructure.

In those scenarios, global configurations need to be passed to the initializer(api.Init*) directly and explicitly and use the configuration format recommended by the framework itself.

PR #162, opened 2020-06-18
Feat/set default config

gorexlv:

#161 

### Special notes for reviews

### DIFF
--- a/core/config/entity.go
+++ b/core/config/entity.go
@@ -105,3 +105,40 @@ func checkValid(conf *SentinelConfig) error {
 	}
 	return nil
 }
+
+func (entity *Entity) AppName() string {
+	return entity.Sentinel.App.Name
+}
+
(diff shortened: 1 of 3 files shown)


louyuting:
Hi, could you please rebase the newest code?

core/config/entity.go
```diff
@@ -105,3 +105,40 @@ func checkValid(conf *SentinelConfig) error {
 	}
 	return nil
 }
+
+func (entity Entity) AppName() string {
```
louyuting, line 109: Would using a pointer receiver make more sense?

gorexlv: If the function logic does not modify the Entity,  it does not seem necessary to use pointers.

louyuting: In future versions, scenarios might exist that read configurations in the slot chain.
This copy operation is not acceptable.

PR #172, opened 2020-06-20
Polish comments and logs in hotspot module

louyuting:
### Special notes for reviews

### DIFF
--- a/core/hotspot/rule.go
+++ b/core/hotspot/rule.go
@@ -79,10 +79,10 @@ type SpecificValue struct {
 }
 
 func (s *SpecificValue) String() string {
-	return fmt.Sprintf("SpecificValue:[ValKind: %+v, ValStr: %s]", s.ValKind, s.ValStr)
+	return fmt.Sprintf("SpecificValue: [ValKind: %+v, ValStr: %s]", s.ValKind, s.ValStr)
 }
 
-// Rule represents the frequency parameter flow control rule
+// Rule represents the hotspot(frequent) parameter flow control rule
 type Rule struct {
 	// Id is the unique id
 	Id string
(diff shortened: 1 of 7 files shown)


core/hotspot/traffic_shaping.go
```diff
@@ -86,13 +86,13 @@ func (c *baseTrafficShapingController) performCheckingForConcurrencyMetric(arg i
 		if concurrency <= specificConcurrency {
 			return nil
 		}
-		return base.NewTokenResultBlocked(base.BlockTypeHotSpotParamFlow, fmt.Sprintf("Frequency params traffic shaping controller, current concurrency: %d, specific concurrency: %d", concurrency, specificConcurrency))
+		return base.NewTokenResultBlocked(base.BlockTypeHotSpotParamFlow, fmt.Sprintf("Hotspot parameters traffic shaping controller, current concurrency: %d, specific concurrency: %d", concurrency, specificConcurrency))
```
sczyh30, line 89: Actually we could carry the rule and the triggered value in `BlockError` (using `NewTokenResultBlockedWithCause`), rather than in the unstructured message. We may improve this later.

core/hotspot/rule.go
```diff
@@ -93,12 +93,19 @@ type Rule struct {
 	// ParamIndex is the index in context arguments slice.
 	ParamIndex int
 	Threshold  float64
-	// MaxQueueingTimeMs is the max queueing time in Throttling ControlBehavior
+	// MaxQueueingTimeMs is the max queueing time
```
sczyh30, line 96: This does not make sense :)

louyuting: Removed

core/hotspot/traffic_shaping.go
louyuting: Updated

sczyh30:
Thanks! We may need to standardize the message and triggered value of `BlockError` later.

PR #176, opened 2020-06-27
Optimize load rules for max size rule

louyuting:
### Special notes for reviews

### DIFF
--- a/core/hotspot/rule_manager.go
+++ b/core/hotspot/rule_manager.go
@@ -6,6 +6,7 @@ import (
 	"sync"
 
 	"github.com/alibaba/sentinel-golang/logging"
+	"github.com/alibaba/sentinel-golang/util"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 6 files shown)


louyuting:
@sczyh30  
This PR optimize the granularity of locks when LoadRules.

- Keep the lock granularity to a minimum
- Reduce memory allocates during write lock holding(avoid to grow map or slice)

But there are still some issues to discuss：

- If the size of rules exceed 10k, the time read lock will cost 10ms+, it also causes cpu and memory jitter.  

Some of my thinking is to sharding and updating the rule list, but this will bring some issues for reuse statistics.

Do you have more suggestion?

tests/maxsize_rule_list_benchmark_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"testing"
+
+	cb "github.com/alibaba/sentinel-golang/core/circuitbreaker"
+)
+
+func Test_Size_1000_Circuit_Breaker_Rules_Update(t *testing.T) {
+	rs := make([]cb.Rule, 0, 1000)
+	intervals := []uint32{10000, 15000, 20000, 25000, 30000}
+	for i := 0; i < 1000; i++ {
+		retryTimeout := intervals[rand.Int()%5]
```
cncal, line 15: `rand.Int()` will generate the same pseudo-random sequence if you don't call `rand.Seed()` first.

louyuting: Updated.

core/hotspot/rule_manager.go
```diff
@@ -98,13 +99,79 @@ func onRuleUpdate(rules []*Rule) (err error) {
...
+	for res, rules := range newRuleMap {
+		m[res] = make([]TrafficShapingController, 0, len(rules))
+	}
+
+	start := util.CurrentTimeNano()
 	tcMux.Lock()
-	defer tcMux.Unlock()
+	defer func() {
+		tcMux.Unlock()
+		logger.Infof("Updating hotspot rule spends %d ns.", util.CurrentTimeNano()-start)
```
sczyh30, line 126: Make it "debug" level?

core/hotspot/rule_manager.go
```diff
@@ -98,13 +99,79 @@ func onRuleUpdate(rules []*Rule) (err error) {
...
+		logRuleUpdate(m)
```
sczyh30, line 127: Here we may need to handle cases regarding panic outside (where the log is not required).

sczyh30:
Thanks!

PR #179, opened 2020-07-03
Refactor consul datasource and polish related code

louyuting:
1. refactor consul datasource
2. add example for datasource: etcdv3, consul

### Special notes for reviews

### DIFF
--- a/ext/datasource/consul/consul.go
+++ b/ext/datasource/consul/consul.go
@@ -13,8 +13,6 @@ import (
 
 type consulDataSource struct {
 	datasource.Base
-	*options
-
 	propertyKey   string
 	kvQuerier     KVQuerier
 	isInitialized util.AtomicBool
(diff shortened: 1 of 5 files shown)


ext/datasource/consul/consul.go
```diff
@@ -77,24 +73,23 @@ func (c *consulDataSource) ReadSource() ([]byte, error) {
 
 	c.queryOptions.WaitIndex = meta.LastIndex
 	if pair == nil {
-		return []byte{}, nil
+		return nil, perrors.New("the kv pair don't exist.")
```
sczyh30, line 76: Maybe non-existence is not actually an error?

example/datasource/consul/main.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/alibaba/sentinel-golang/ext/datasource"
+	"github.com/alibaba/sentinel-golang/ext/datasource/consul"
+	"github.com/hashicorp/consul/api"
+)
+
+func main() {
+	ds := startConsulDs()
+
+	scanner := bufio.NewScanner(os.Stdin)
```
sczyh30, line 16: Here we may make it a real demo (with `sentinel.Entry(xxx)`) so that users could observe the effect as soon as the rules have been modified.

sczyh30:
And we may also add some logs (when the data-source has been initialized) so that users could identify which key is being watched.

ext/datasource/consul/consul.go
```diff
@@ -25,14 +23,14 @@ type consulDataSource struct {
 var (
 	ErrNilConsulClient     = errors.New("nil consul client")
 	ErrInvalidConsulConfig = errors.New("invalid consul config")
+	ErrKeyIsNotExisted     = errors.New("key is not existed now")
```
sczyh30, line 26: does not exit

sczyh30:
Thanks!

PR #187, opened 2020-07-23
Improve all rule checking slots with standard BlockError representation

sczyh30:

Improve all rule checking slots with standard `BlockError` representation.

NONE

### Describe how you did it

This PR follows a simple BlockError specification:

- The `Error()` SHOULD start with `SentinelBlockError: ${blockType}`. If block message is provided, append it to the `Error()` value. The `blockMsg` is not required.
- It's recommended to carry the triggered rule and the "snapshot" value in the `BlockError`.

This PR improves all rule checking slots with more standard `BlockError` representation.

### Describe how to verify it

Run the test cases.

### Special notes for reviews

NONE

### DIFF
--- a/core/flow/tc_throttling.go
+++ b/core/flow/tc_throttling.go
@@ -34,7 +34,7 @@ func (c *ThrottlingChecker) DoCheck(_ base.StatNode, acquireCount uint32, thresh
 		return nil
 	}
 	if threshold <= 0 {
-		return base.NewTokenResultBlocked(base.BlockTypeFlow, "Flow")
+		return base.NewTokenResultBlocked(base.BlockTypeFlow)
 	}
 	// Here we use nanosecond so that we could control the queueing time more accurately.
 	curNano := util.CurrentTimeNano()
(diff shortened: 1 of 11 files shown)


core/hotspot/traffic_shaping.go
```diff
@@ -87,16 +87,14 @@ func (c *baseTrafficShapingController) performCheckingForConcurrencyMetric(arg i
 			return nil
 		}
 		return base.NewTokenResultBlockedWithCause(base.BlockTypeHotSpotParamFlow,
-			fmt.Sprintf("Blocked by specific concurrency threshold, arg: %+v, current concurrency: %d,  specific concurrency: %d", arg, concurrency, specificConcurrency),
-			c.BoundRule(), concurrency)
+			fmt.Sprintf("arg=%v", arg), c.BoundRule(), nil)
```
louyuting, line 90: Maybe we should carry the snapshot?

core/flow/tc_throttling.go
```diff
@@ -34,7 +34,7 @@ func (c *ThrottlingChecker) DoCheck(_ base.StatNode, acquireCount uint32, thresh
 		return nil
 	}
 	if threshold <= 0 {
-		return base.NewTokenResultBlocked(base.BlockTypeFlow, "Flow")
+		return base.NewTokenResultBlocked(base.BlockTypeFlow)
```
louyuting, line 37: Maybe we need to carry the triggered rule?

core/hotspot/traffic_shaping.go
sczyh30: Updated.

core/flow/tc_throttling.go
sczyh30: Yes, the design regarding tc_throttling.go should be improved.

PR #197, opened 2020-07-31
Optimize the way to initialize the start time of the AtomicBucketWrapArray

louyuting:
Make the start time of AtomicBucketWrapArray align to standard time.
Fix the typo in NewLeapArray.

### Special notes for reviews

### DIFF
--- a/core/stat/base/leap_array.go
+++ b/core/stat/base/leap_array.go
@@ -49,34 +49,48 @@ type AtomicBucketWrapArray struct {
 	data   []*BucketWrap
 }
 
-// New AtomicBucketWrapArray with initializing field data
-// Default, automatically initialize each BucketWrap
-// len: length of array
-// bucketLengthInMs: bucket length of BucketWrap
-// generator: generator to generate bucket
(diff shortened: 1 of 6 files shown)


core/stat/base/leap_array.go
```diff
@@ -49,28 +49,25 @@ type AtomicBucketWrapArray struct {
...
 		length: len,
 		data:   make([]*BucketWrap, len),
 	}
 
 	// automatically initialize each BucketWrap
 	// tail BucketWrap of data is initialized with current time
-	startTime := calculateStartTime(util.CurrentTimeMillis(), bucketLengthInMs)
-	for i := len - 1; i >= 0; i-- {
+	interval := uint64(bucketLengthInMs * uint32(len))
+	arrayStartTime := now - (now % interval)
```
sczyh30, line 61: The startTime should be `now - now % bucketLengthInMs`?

For example (in related test cases), sampleCount=10, bucketLength=200ms, now=1596199317921, then the first (and oldest) bucket of created data structure should be 1596199317800

sczyh30:
Thanks!

PR #201, opened 2020-08-11
Fix the inveracious logging.Logger to customize.

louyuting:
In logging module, current implementation cannot customize Logger implementation by user.
Current exported function, logging.ResetDefaultLogger(), is not working as expected. 

### Special notes for reviews

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -29,7 +29,7 @@ const (
 var (
 	globalLogLevel = Info
 
-	defaultLogger = NewConsoleLogger(DefaultNamespace)
+	globalLogger = NewConsoleLogger(DefaultNamespace)
 )
 
 func GetGlobalLoggerLevel() Level {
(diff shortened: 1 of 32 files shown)


logging/logging.go
```diff
@@ -39,33 +40,33 @@ func SetGlobalLoggerLevel(l Level) {
 	globalLogLevel = l
 }
 
-func GetDefaultLogger() Logger {
-	return defaultLogger
+func GetGlobalLogger() Logger {
```
sczyh30, line 43: Actually we may export functions like `Infof`, `Warn` as delegates to the default logger in `logging` package, so it could be more convenient to use:

```go
logging.Infof(xxx)
```

eddycjy:
Thanks for the fix, I think it is basically available.

sczyh30:
Nice work. Thanks!

PR #205, opened 2020-08-17
Using circuitbreaker.Rule to define slowRequest/errorRatio/errorCount Rule

louyuting:
Currently, there are three rules to define circuit breaker rule: slowRequest/errorRatio/errorCount Rule.
The rule definition of circuit breaker rule is complex and there are also limitations to scalability.
So this PR will consolidate these circuit breaker rules to circuitbreaker.Rule.

### Special notes for reviews


core/circuitbreaker/rule.go
```diff
@@ -35,307 +33,93 @@ func (s Strategy) String() string {
...
-	return b.StatIntervalMs
+func (r *Rule) String() string {
+	// fallback string
+	return fmt.Sprintf("{id=%s,resource=%s, strategy=%s, RetryTimeoutMs=%d, MinRequestAmount=%d, StatIntervalMs=%d, MaxAllowedRtMs=%d, Threshold=%f}",
+		r.Id, r.Resource, r.Strategy, r.RetryTimeoutMs, r.MinRequestAmount, r.StatIntervalMs, r.MaxAllowedRtMs, r.Threshold)
 }
 
-func (b *RuleBase) IsApplicable() error {
-	if len(b.Resource) == 0 {
+func (r *Rule) isApplicable() error {
```
louyuting, line 69: Export this function is user-friendly?
How do you think of it?

sczyh30: Since the rule itself is not an interface anymore, maybe we could make it an exported util checking function (like `IsValidFlowRule` in `flow/rule_manager.go`) instead of a member function?

louyuting: make sense

core/circuitbreaker/rule_manager.go
```diff
@@ -315,3 +312,25 @@ func RemoveCircuitBreakerGenerator(s Strategy) error {
...
+	if int(r.Strategy) < int(SlowRequestRatio) || int(r.Strategy) > int(ErrorCount) {
+		return errors.New("invalid Strategy")
+	}
+	if r.StatIntervalMs <= 0 {
+		return errors.New("invalid StatIntervalMs")
+	}
+	if r.Threshold < 0 {
+		return errors.New("invalid Threshold")
+	}
+	if r.Strategy == SlowRequestRatio && (r.Threshold < 0.0 || r.Threshold > 1.0) {
```
sczyh30, line 329: `r.Threshold < 0.0` is not needed as it has been checked before.

core/circuitbreaker/rule_manager.go
```diff
@@ -315,3 +312,25 @@ func RemoveCircuitBreakerGenerator(s Strategy) error {
...
+		return errors.New("invalid slow request ratio threshold (valid range: [0.0, 1.0])")
+	}
+	if r.Strategy == ErrorRatio && (r.Threshold < 0.0 || r.Threshold > 1.0) {
```
sczyh30, line 332: Ditto

core/circuitbreaker/rule_manager.go
```diff
@@ -315,3 +312,25 @@ func RemoveCircuitBreakerGenerator(s Strategy) error {
...
 }
+
+func IsValid(r *Rule) error {
+	if len(r.Resource) == 0 {
+		return errors.New("empty resource name")
+	}
+	if int(r.Strategy) < int(SlowRequestRatio) || int(r.Strategy) > int(ErrorCount) {
+		return errors.New("invalid Strategy")
+	}
+	if r.StatIntervalMs <= 0 {
```
sczyh30, line 323: retryTimeout should also be positive

core/circuitbreaker/rule_manager.go
louyuting: updated

core/circuitbreaker/rule_manager.go
louyuting: updated

core/circuitbreaker/rule_manager.go
louyuting: updated

sczyh30 (approved):
LGTM

sczyh30:
Nice work. Thanks!

PR #202, opened 2020-08-12
fix circuit breaker half-open state issue.

louyuting:
### Describe what this PR does / why we need it
Fix the issue: #196 
add defer to handle panic in entry.Exit func and move context recovery logic to defer

### Does this pull request fix one issue?
#196 

### Describe how to verify it
UT and integration test
tests/core/circuitbreaker/circuitbreaker_slot_integration_test.go 
### Special notes for reviews

PR #206, opened 2020-08-17
Refine api

louyuting:

Currently, api.Entry() may return (nil, blockError).
Return nil is not user-friendly, normally, the user might use defer to perform Exit func, that causes panic.
Actually, we had done work to guarantee to exec Exit function exactly once.
From me perspective, I think to return entry is okay.

### Special notes for reviews

### DIFF
--- a/core/base/entry.go
+++ b/core/base/entry.go
@@ -29,8 +29,11 @@ func (e *SentinelEntry) SetError(err error) {
 	}
 }
 
-func (e *SentinelEntry) Context() *EntryContext {
-	return e.ctx
+func (e *SentinelEntry) Err() error {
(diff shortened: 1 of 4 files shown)


core/base/entry.go
```diff
@@ -29,8 +29,11 @@ func (e *SentinelEntry) SetError(err error) {
 	}
 }
 
-func (e *SentinelEntry) Context() *EntryContext {
```
louyuting, line 32: From my perspective, export Context don't make sense.
1. Context should be sentinel internal status,
2. Context is maintained by pool, exporting might cause fatal bug.

louyuting:
Close B/C #202  had covered here.

PR #218, opened 2020-08-31
fix bug: incorrectly defined storeToken using uint in warmup flow control

louyuting:
https://github.com/alibaba/sentinel-golang/blob/36a09c509daba8e3d790157f55af7587b688d39b/core/flow/tc_warm_up.go#L21

Define storedTokens as uint64 might cause bug.
https://github.com/alibaba/sentinel-golang/blob/36a09c509daba8e3d790157f55af7587b688d39b/core/flow/tc_warm_up.go#L78

### Special notes for reviews

### DIFF
--- a/core/flow/tc_warm_up.go
+++ b/core/flow/tc_warm_up.go
@@ -18,14 +18,14 @@ type WarmUpTrafficShapingCalculator struct {
 	warningToken      uint64
 	maxToken          uint64
 	slope             float64
-	storedTokens      *uint64
-	lastFilledTime    *uint64
+	storedTokens      int64
+	lastFilledTime    uint64
 }
 
 func NewWarmUpTrafficShapingCalculator(rule *FlowRule) *WarmUpTrafficShapingCalculator {
-	if rule.WarmUpColdFactor == 0 {
+	if rule.WarmUpColdFactor <= 1 {
 		rule.WarmUpColdFactor = config.DefaultWarmUpColdFactor
-		logging.Warnf("[NewWarmUpTrafficShapingCalculator] No set WarmUpColdFactor,use default values: %d", config.DefaultWarmUpColdFactor)
+		logging.Warnf("[NewWarmUpTrafficShapingCalculator] invalid WarmUpColdFactor,use default values: %d", config.DefaultWarmUpColdFactor)
 	}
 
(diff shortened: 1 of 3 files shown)


core/flow/tc_warm_up.go
```diff
@@ -41,64 +41,67 @@ func NewWarmUpTrafficShapingCalculator(rule *FlowRule) *WarmUpTrafficShapingCalc
 		maxToken:          maxToken,
 		slope:             slope,
 		threshold:         rule.Count,
-		storedTokens:      new(uint64),
-		lastFilledTime:    new(uint64),
+		storedTokens:      0,
```
sanxun0325, line 44: The default value is 0 and no assignment is required。

core/flow/tc_warm_up.go
```diff
@@ -41,64 +41,67 @@ func NewWarmUpTrafficShapingCalculator(rule *FlowRule) *WarmUpTrafficShapingCalc
...
 	previousQps := node.GetPreviousQPS(base.MetricEventPass)
-	d.syncToken(previousQps)
+	c.syncToken(previousQps)
 
-	restToken := atomic.LoadUint64(d.storedTokens)
-	if restToken >= d.warningToken {
-		aboveToken := restToken - d.warningToken
-		warningQps := math.Nextafter(1.0/(float64(aboveToken)*d.slope+1.0/d.threshold), math.MaxFloat64)
+	restToken := atomic.LoadInt64(&c.storedTokens)
+	if restToken < 0 {
```
sanxun0325, line 56: There is no situation where the restToken is less than 0.

core/flow/tc_warm_up.go
louyuting: Yes, just an explicit declaration

core/flow/tc_warm_up.go
louyuting: It might have;
```
if atomic.CompareAndSwapInt64(&c.storedTokens, oldValue, newValue) {
		if currentValue := atomic.AddUint64(d.storedTokens, uint64(0-passQps)); currentValue < 0 {			if currentValue := atomic.AddInt64(&c.storedTokens, int64(-passQps)); currentValue < 0 {
```
In concurrency, after `currentValue := atomic.AddInt64(&c.storedTokens, int64(-passQps)); ` might cause  storedTokens is less than 0;

sczyh30 (approved):
LGTM

sczyh30:
Thanks!

PR #224, opened 2020-09-04
Refactoring: Change NewBlockErrorFromDeepCopy behavior.

sdttttt:

In Golang, when the parameter is a struct of non-pointer type, the input parameter will make a memory copy.
on this function, I added tests.

### Describe how to verify it

Run the `TestNewBlockErrorFromDeepCopy`. in `block_error_test.go`.

### Special notes for reviews


sdttttt:
There's something wrong with my branch.

This PR needs to be closed.

PR #226, opened 2020-09-06
Refactoring: Change NewBlockErrorFromDeepCopy behavior.

sdttttt:

In Golang, when the parameter is a struct of non-pointer type, the input parameter will make a memory copy.
on this function, I added tests.

### Describe how to verify it

Run the `TestNewBlockErrorFromDeepCopy`. in `block_error_test.go`.

### Special notes for reviews

#224 fixed.


core/base/block_error.go
```diff
@@ -29,13 +29,8 @@ func (e *BlockError) TriggeredValue() interface{} {
...
 }
 
-func NewBlockErrorFromDeepCopy(from *BlockError) *BlockError {
-	return &BlockError{
-		blockType:     from.blockType,
-		blockMsg:      from.blockMsg,
-		rule:          from.rule,
-		snapshotValue: from.snapshotValue,
-	}
+func NewBlockErrorFromDeepCopy(blockErr BlockError) *BlockError {
```
louyuting, line 32: Maybe copy explicitly is better?

sdttttt: I think maybe we can get better performance by having the compiler do this for us.
this can be confusing on grammatically.

PR #231, opened 2020-09-07
Refine API and exported structure naming in core pkg

louyuting:
Refine API and exported structure naming in core pkg

### Special notes for reviews


api/doc.go
```diff
@@ -0,0 +1,51 @@
+// Package api is the topmost API interface package for users using sentinel-golang(aka sentinel)
+//
+// Users must initialize sentinel before loading sentinel rules. Sentinel support three ways to perform initialization:
```
sczyh30, line 3: sentinel -> Sentinel (in all cases as a brand)

api/doc.go
```diff
@@ -0,0 +1,51 @@
+// Package api is the topmost API interface package for users using sentinel-golang(aka sentinel)
```
sczyh30, line 1: The first sentence should end with `.`

api/doc.go
```diff
@@ -0,0 +1,51 @@
...
+// Here is the example code to use sentinel:
+//
+//  import sentinel "github.com/alibaba/sentinel-golang/api"
+//
+//  err := sentinel.InitDefault()
+//  if err != nil {
+//      log.Fatal(err)
+//  }
+//
+//  //Load sentinel rules
```
sczyh30, line 18: The example might be too complicated? Actually we may introduce APIs in this package.

sczyh30 (approved):

sczyh30:
Thanks!

PR #237, opened 2020-09-10
Split Independent mod for adapter 

louyuting:
Split Independent mod for adapter and clean sentinel mod

### Special notes for reviews


louyuting:
After investigation, split independent repo maybe better,

PR #238, opened 2020-09-12
Improve Config API

louyuting:
Improve sentinel initialization mechanism

1. The logging module init function sets the default logger to be a file logger(sentinel-record.log). Even if the user does not initialize sentinel, it will work properly.
2. Reset logger by configEntity/yaml/default is optional. Reset logger only if the user sets logger in configuration.

None

### Describe how to verify it
UT

### Special notes for reviews
@sczyh30

### DIFF
--- a/api/init.go
+++ b/api/init.go
@@ -9,6 +9,10 @@ import (
 	"github.com/alibaba/sentinel-golang/util"
 )
 
+// Initialization func initialize the Sentinel's runtime environment, including:
+// 		1. override global config, from manually config or yaml file or env variable
+//		2. override global logger
+// 		3. initiate core component async task, including: metric log, system statistic...
 // InitDefault initializes Sentinel using the configuration from system
 // environment and the default value.
 func InitDefault() error {
@@ -47,11 +51,16 @@ func InitWithConfigFile(configPath string) error {
 // initCoreComponents init core components with default config
 // it's better SetDefaultConfig before initCoreComponents
 func initCoreComponents() error {
-	if err := metric.InitTask(); err != nil {
-		return err
+	if config.MetricLogFlushIntervalSec() > 0 {
+		if err := metric.InitTask(); err != nil {
+			return err
+		}
 	}
 
-	system.InitCollector(config.SystemStatCollectIntervalMs())
+	if config.SystemStatCollectIntervalMs() > 0 {
+		system.InitCollector(config.SystemStatCollectIntervalMs())
+	}
+
 	if config.UseCacheTime() {
 		util.StartTimeTicker()
 	}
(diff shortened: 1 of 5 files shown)


api/init.go
```diff
@@ -46,12 +50,17 @@ func InitWithConfigFile(configPath string) error {
 
 // initCoreComponents init core components with default config
 // it's better SetDefaultConfig before initCoreComponents
-func initCoreComponents() error {
-	if err := metric.InitTask(); err != nil {
-		return err
+func initCoreComponents(confEntity *config.Entity) error {
+	if confEntity.EnableMetricLog() {
```
sczyh30, line 54: > Add field EnableMetricLog in config.Entity to indicate whether run metric log task, default false;

This can be achieved when `MetricLogFlushIntervalSec=0`.

> Add field EnableSystemStat in config.Entity to indicate whether run system statistic task, default false;

This can be achieved when `SystemStatCollectIntervalMs=0`.

louyuting: Make sense. the default value of `MetricLogFlushIntervalSec` and `SystemStatCollectIntervalMs` is zero, that meaning not run metric log task and system statistic task.
Right?

louyuting: Updated

sczyh30: > Make sense. the default value of `MetricLogFlushIntervalSec` and `SystemStatCollectIntervalMs` is zero, that meaning not run metric log task and system statistic task.
> Right?

DefaultMetricLogFlushIntervalSec   uint32 = 1
DefaultSystemStatCollectIntervalMs uint32 = 1000

louyuting: Sorry for misunderstanding.  It should be "the setting value of MetricLogFlushIntervalSec and SystemStatCollectIntervalMs is zero, that meaning not run metric log task and system statistic task"..

logging/logging.go
```diff
@@ -20,12 +21,21 @@ const (
...
+	DefaultNamespace = "sentinel-go"
 	// RecordLogFileName represents the default file name of the record log.
 	RecordLogFileName = "sentinel-record.log"
 	DefaultDirName    = "logs" + string(os.PathSeparator) + "csp" + string(os.PathSeparator)
 )
 
+func init() {
+	defaultFileLogger, err := NewSimpleFileLogger(GetDefaultLogDir()+RecordLogFileName, DefaultNamespace, 0)
+	if err != nil {
+		globalLogger.Error("fail to init global file logger")
```
sczyh30, line 33: Carry with error message?

louyuting: Sure, updated

sczyh30 (approved):
LGTM

sczyh30 (approved):

sczyh30:
Thanks!

PR #240, opened 2020-09-12
Fix CI accidental failure

louyuting:
Because of golangci-lint's default timeout is 1m. It might cause CI failure.
 

### Special notes for reviews


sczyh30 (approved):

sczyh30:
Thanks!

PR #245, opened 2020-09-21
Rule Manager: circuit breaker module supports resource level replacement

sanxun0325:
circuit breaker module supports resource level replacement

### Special notes for reviews


sanxun0325:
Related pr:https://github.com/alibaba/sentinel-golang/pull/374

PR #247, opened 2020-09-22
split isolation pkg to implement concurrency traffic control

louyuting:
split isolation pkg to implement concurrency traffic control

### Special notes for reviews

### DIFF
--- a/core/isolation/rule_manager.go
+++ b/core/isolation/rule_manager.go
@@ -14,6 +14,7 @@ var (
 	rwMux   = &sync.RWMutex{}
 )
 
+// LoadRules loads the given isolation rules to the rule manager, while all previous rules will be replaced.
 func LoadRules(rules []*Rule) (updated bool, err error) {
 	updated = true
 	err = nil
@@ -42,11 +43,14 @@ func LoadRules(rules []*Rule) (updated bool, err error) {
 	return
 }
 
+// ClearRules clears all the rules in isolation module.
 func ClearRules() error {
 	_, err := LoadRules(nil)
 	return err
 }
 
+// GetRules returns all the rules based on copy.
+// It doesn't take effect for isolation module if user changes the rule.
 func GetRules() []Rule {
 	rules := getRules()
 	ret := make([]Rule, 0, len(rules))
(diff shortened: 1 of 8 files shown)


core/isolation/rule_manager.go
```diff
@@ -0,0 +1,136 @@
...
+	ruleMap = m
+	return
+}
+
+func ClearRules() error {
+	_, err := LoadRules(nil)
+	return err
+}
+
+func GetRules() []Rule {
```
louyuting, line 50: I  will add comments later

louyuting: Updated

core/isolation/slot.go
```diff
@@ -0,0 +1,47 @@
...
+			} else {
+				curCount = 0
+				logging.Error(errors.New("negative concurrency"), "", "rule", rule)
+			}
+			if curCount+acquireCount > threshold {
+				return false, rule, curCount
+			}
+		}
+	}
+	return true, nil, curCount
```
sczyh30, line 46: Maybe we could also carry the rule in normal result, or add comments about it (nil when passed).

core/isolation/rule_manager.go
```diff
@@ -0,0 +1,147 @@
...
+// getRulesOfResource is an internal interface.
+func getRulesOfResource(res string) []*Rule {
+	rwMux.RLock()
+	defer rwMux.RUnlock()
+
+	resRules, exist := ruleMap[res]
+	if !exist {
+		return nil
+	}
+	ret := make([]*Rule, 0, len(resRules))
```
sanxun0325, line 93: Using copy Func may be more concise

core/isolation/rule_manager.go
```diff
@@ -0,0 +1,147 @@
...
+
+var (
+	ruleMap = make(map[string][]*Rule)
+	rwMux   = &sync.RWMutex{}
+)
+
+// LoadRules loads the given isolation rules to the rule manager, while all previous rules will be replaced.
+func LoadRules(rules []*Rule) (updated bool, err error) {
+	updated = true
+	err = nil
```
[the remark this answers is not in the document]
louyuting: It just keep API consistency with other module

core/isolation/rule_manager.go
louyuting: Updated

core/isolation/slot.go
louyuting: Comments had been added

sczyh30 (approved):
LGTM

Issue #248, opened 2020-09-22
Rule Manager: Make LoadRule() serialization  and introduce recover logic when panic for LoadRules in core module


louyuting:
## Issue Description

Type: *feature request*

### Describe what feature you want
Currently, the logic of `LoadRules` is below:
1. acquire Write Lock
2. rebuild traffic shaping control from new rules
3. update core module's tcMap
4. release Write Lock
5. defer test recover().

There are two question:
1. The Write Lock take too long
2. If occurs panic when LoadRules, the origin tcMap could not be rollback.

I have a proposal to optimize this issue: (for example: circuit breaker module)
1. Define a channel(no buffer) in circuit breaker  manager, LoadRules/ClearRule/(other write operation) push rules into channel
. It guarantee the single thread to write.
2. init function of  circuit breaker module run a goroutine to receive channel and do Write operation.
3. Write operation copy the origin tcMap(need write lock), and do load logic. If any panic occurs, rollback to copied tcMap.

### Additional context

Scope

- [ ] flow

- [ ] circuit breaker

- [ ] isolation

- [x] hotpot

- [ ] system

sanxun0325:
@i will solve it@

liqiangz:
I will solve it

louyuting:
> I will solve it

Welcomed~

louyuting:
refer to #354

PR #251, opened 2020-09-23
Rule Manager: Make LoadRule() serialization and introduce recover logic when panic for LoadRules in circuit breaker module

sanxun0325:
Enhance mutex and recover logic for LoadRules function in core module.
Reduce the granularity of locks

Fix:https://github.com/alibaba/sentinel-golang/issues/248

### Special notes for reviews


sanxun0325:
Related pr https://github.com/alibaba/sentinel-golang/pull/354/files

PR #252, opened 2020-09-24
hotspot module support specific threshold for arbitrary kind specific parameter

louyuting:
hotspot module support specific threshold for arbitrary kind specific parameter

### Special notes for reviews


ext/datasource/hotspot_rule_converter.go
```diff
@@ -0,0 +1,98 @@
...
+	case KindFloat64:
+		return "KindFloat64"
+	default:
+		return "Undefined"
+	}
+}
+
+// SpecificValue indicates the specific param, contain the supported param kind and concrete value.
+type SpecificValue struct {
+	ValKind   ParamKind `json:"valKind"`
```
sanxun0325, line 45: Can you verify valkind using https://github.com/go-playground/validator

louyuting: The idea is well, From my perspective, I think it does more harm than good.
Sentinel, as a base component, is not a good choice for introducing too many dependent packages. We should keep package dependencies pure.

sczyh30:
Nice work. Thanks!

Issue #111, opened 2020-03-23
Add value cache/checking mechanism for rule managers


louyuting:
## Issue Description

Type: *feature request*

### Describe what feature you want
In order to avoid the meaningless updates to the property, downstream of property manager should 
cache last update value to check consistency.

hellosmallstone:
I will try to implement this function.
Please assign this to me.

louyuting:
Contributions are welcomed

louyuting:
@hellosmallstone  Hi, any process?

sanxun0325:
This issue can be assigned to me! I will resolve it.

louyuting:
Welcomed~

sczyh30:
Resolved via #253, #321, #322, #324. Cheers! 🚀 
We may need a universal design after 1.0 GA.

PR #253, opened 2020-09-24
[ISSUES 111]Flow rule cache/check

sanxun0325:

Fix:https://github.com/alibaba/sentinel-golang/issues/111
### Describe how you did it

### Describe how to verify it
ut

### Special notes for reviews


core/flow/rule_manager.go
```diff
@@ -88,12 +89,16 @@ func onRuleUpdate(rules []*Rule) (err error) {
...
 	tcMap = m
+	currentRules = rules
 	return nil
 }
 
 // LoadRules loads the given flow rules to the rule manager, while all previous rules will be replaced.
 func LoadRules(rules []*Rule) (bool, error) {
 	// TODO: rethink the design
+	if isEqual := reflect.DeepEqual(currentRules, rules); isEqual {
+		return isEqual, errors.New("The current Rules is the same as the rules to be loaded")
```
louyuting, line 100: it should return `false, nil`

louyuting:
@sczyh30  How do you think about this design?

core/flow/rule_manager.go
sanxun0325: updated

louyuting:
Could you please solve conflicts?

PR #256, opened 2020-09-27
enhance refreshable_file: watch source file event after remove/rename

weiyuanke:
fix refreshable_file

### Special notes for reviews


louyuting (approved):
Nice work.
Thanks for contribution

PR #261, opened 2020-10-03
Fix error message of unit test for bucket wrapper size

luckyxiaoqiang:
Fix error message of unit test for bucket wrapper size.

### Special notes for reviews


louyuting (approved):
Thanks for contribution

PR #265, opened 2020-10-04
Use path/filepath pkg to do file path related operations

luckyxiaoqiang:
Use path/filepath pkg to do file path related operations.  Package filepath implements utility routines for manipulating filename paths in a way compatible with the target operating system-defined file paths. The filepath package uses either forward slashes or backslashes, depending on the operating system.

### Special notes for reviews


louyuting (approved):
Thanks for contribution

PR #268, opened 2020-10-07
Flow rule cache/check

sanxun0325:
Describe what this PR does / why we need it

Does this pull request fix one issue?
Fix:#111

Describe how you did it

Describe how to verify it
ut

Special notes for reviews

### DIFF
--- a/core/flow/rule_manager.go
+++ b/core/flow/rule_manager.go
@@ -2,6 +2,7 @@ package flow
 
 import (
 	"fmt"
+	"reflect"
 	"sync"
 
 	"github.com/alibaba/sentinel-golang/core/base"
@@ -34,6 +35,7 @@ var (
 		readOnlyMetric:    base.NopReadStat(),
 		writeOnlyMetric:   base.NopWriteStat(),
 	}
+	currentRules = make([]*Rule, 0)
 )
 
 func init() {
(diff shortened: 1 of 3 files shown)


core/flow/rule_manager_test.go
```diff
@@ -373,3 +373,27 @@ func Test_buildRulesOfRes(t *testing.T) {
...
+				Resource:               "some-test",
+				Threshold:              10,
+				TokenCalculateStrategy: Direct,
+				ControlBehavior:        Reject,
+			},
+		})
+		assert.Nil(t, err)
+		assert.False(t, ok)
+	})
+}
```
luckyxiaoqiang, line 399: How about add assert that `LoadRules` does return `true` when rules updated.

core/flow/rule_manager.go
```diff
@@ -28,6 +29,7 @@ var (
 	tcGenFuncMap = make(map[trafficControllerGenKey]TrafficControllerGenFunc)
 	tcMap        = make(TrafficControllerMap)
 	tcMux        = new(sync.RWMutex)
+	currentRules = make([]*Rule, 0)
```
luckyxiaoqiang, line 32: Better to protect it by a lock to avoid concurrent read and write. And should we consider the `Rule` order of different resources? How about a map?

sanxun0325: Locks are used in onRuleUpdate func,don't have to think about different resources, because want to cache global rules.

core/flow/rule_manager_test.go
sanxun0325: Other ut has done a lot of validation for cases that return true

core/flow/rule_manager.go
luckyxiaoqiang: `LoadRules` reads `currentRules ` without lock.

sanxun0325: updated

sczyh30 (approved):

PR #269, opened 2020-10-07
enhance datasource ut

sanxun0325:
add datasource base Handle() ut

### Describe how to verify it
go test ./...

### Special notes for reviews

### DIFF
--- a/ext/datasource/datasource_test.go
+++ b/ext/datasource/datasource_test.go
@@ -4,6 +4,8 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/alibaba/sentinel-golang/core/flow"
+	"github.com/alibaba/sentinel-golang/core/system"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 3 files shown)


ext/datasource/datasource_test.go
```diff
@@ -63,3 +74,24 @@ func TestBase_indexOfHandler(t *testing.T) {
...
+	t.Run("TestBase_handleErr", func(t *testing.T) {
+		b := &Base{
+			handlers: make([]PropertyHandler, 0),
+		}
+		h := NewSystemRulesHandler(SystemRuleJsonArrayParser)
+		b.handlers = append(b.handlers, h)
+		err := b.Handle([]byte(systemRuleErr))
+		assert.NotNil(t, err)
+	})
+}
```
luckyxiaoqiang, line 97: Maybe better to add a case of multi handlers.

sanxun0325: updated

louyuting (approved):
Thanks for your contribution

PR #271, opened 2020-10-09
enhance: Strengthen the workflow.

sdttttt:

1. Split the workflow for `unit testing` and `static checking`. Now the two of them are executed in **parallel**.

2. Fixed that `golangci-lint` does not work well and accurately.Now `golangci-lint` will give an accurate opinion on the revision.

3. Add vulnerability detection (CodeQL) to the static check workflow.

4. Some of the code has been fixed based on `golangci-lint` 's suggested changes.

### Special notes for reviews

> **Note:** the new CI job only **1m 30s**, but the CodeQL job about **4m**.

We can see the proposed changes in the golangci report at GithubActions.


louyuting:
@sdttttt  Could you please fix CI?

sdttttt:
Emmm. I'll try.

Golangci-lint gave quite a number of suggestions for revision.
➢ @sdttttt Could you please fix CI?

sdttttt:
Now, the proposed changes proposed by `golangci-lint` will not affect the success or failure of CI.

sdttttt:
@louyuting PTAL.

sdttttt:
I'm sorry, I can't focus on that anymore.

PR #272, opened 2020-10-10
enhance core stat ut

sanxun0325:
enhance core stat ut

### Describe how to verify it
go test ./...

### Special notes for reviews

### DIFF
--- a/core/stat/base/leap_array_test.go
+++ b/core/stat/base/leap_array_test.go
@@ -6,6 +6,8 @@ import (
 	"testing"
 	"unsafe"
 
+	"github.com/stretchr/testify/assert"
+
 	"github.com/alibaba/sentinel-golang/util"
 	"github.com/stretchr/testify/mock"
 )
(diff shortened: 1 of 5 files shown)


core/stat/base/sliding_window_metric_test.go
```diff
@@ -3,6 +3,8 @@ package base
 import (
 	"testing"
 
+	"github.com/alibaba/sentinel-golang/util"
```
louyuting, line 6: rearrange imports by "go imports"

core/stat/base/leap_array_test.go
```diff
@@ -6,6 +6,8 @@ import (
 	"testing"
 	"unsafe"
 
+	"github.com/stretchr/testify/assert"
```
louyuting, line 9: rearrange imports by "go imports"

sanxun0325: updated

core/stat/base/sliding_window_metric_test.go
sanxun0325: updated

louyuting (approved):
Thanks for contribution

PR #277, opened 2020-10-11
Refine pooled EntryOptions management

luckyxiaoqiang:
Refine pooled EntryOptions management by putting it to pool at where we get it. Otherwise, it would be difficult to track its life cycle and not benefit to future maintenance.

### Special notes for reviews


louyuting (approved):
Thanks for contribution and look forward for more!

PR #288, opened 2020-10-13
Refine string builder usage in logging pkg

luckyxiaoqiang:
Use string builder grow to reduce memory allocations in logging pkg.

Benchmark result:
```
goos: darwin
goarch: amd64
pkg: github.com/alibaba/sentinel-golang/logging
BenchmarkAssembleMsg/assemble_msg_using_string_builder_with_grow-4                742695              1681 ns/op             624 B/op          8 allocs/op
BenchmarkAssembleMsg/assemble_msg_using_string_builder_without_grow-4             678778              1901 ns/op             728 B/op         12 allocs/op
```

### Special notes for reviews

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -27,6 +27,8 @@ const (
 	// RecordLogFileName represents the default file name of the record log.
 	RecordLogFileName = "sentinel-record.log"
 	GlobalCallerDepth = 4
+
+	defaultLogMsgBufferSize = 256
 )
 
 var (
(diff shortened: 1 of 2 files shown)


logging/logging.go
```diff
@@ -31,6 +31,8 @@ const (
 
 var (
 	DefaultDirName = filepath.Join("logs", "csp")
+
+	defaultLogMsgBufferSize = 256
```
louyuting, line 35: maybe make it const is better?

luckyxiaoqiang: A const is not convenient for benchmark test.

louyuting: From my perspective, Benchmark is to test performance and shouldn't impact the code.

luckyxiaoqiang: OK, I'll refine it.

luckyxiaoqiang: OK, I'll refine it.

louyuting (approved):
Thanks for contribution!

PR #289, opened 2020-10-13
Refine metric item stringer

luckyxiaoqiang:
Refine metric item stringer by string builder write methods instead of `fmt.Fprintf`.

Benchmark result before refinement:
```
goos: darwin
goarch: amd64
pkg: github.com/alibaba/sentinel-golang/core/base
BenchmarkMetricItem_ToFatString-4        1245118               966 ns/op             208 B/op         12 allocs/op
```

Benchmark result after refinement:
```
goos: darwin
goarch: amd64
pkg: github.com/alibaba/sentinel-golang/core/base
BenchmarkMetricItem_ToFatString-4        2595044               436 ns/op             176 B/op          3 allocs/op
```

### Special notes for reviews


sczyh30:
Thanks for contributing. This improvement is great, but it may affect the readability, and it's not frequently invoked (usually only per-second). For this kind of cases, maybe elegance counts more?

luckyxiaoqiang:
> Thanks for contributing. This improvement is great, but it may affect the readability, and it's not frequently invoked (usually only per-second). For this kind of cases, maybe elegance counts more?

Thanks for your advice. We really need to do a balance between performance and readability.
According to my understanding, it will be invoked for each resource(usually once per-second). Do we need to consider large number of resources?

PR #300, opened 2020-10-22
Use NOP stat for flow controller that don't need stat

luckyxiaoqiang:
Use NOP stat for flow controller that don't need stat.

### Special notes for reviews


louyuting (approved):
Nice work!

PR #309, opened 2020-10-27
Enhance logging

luckyxiaoqiang:
- Logging supports auto conversion of error interface.
- Refine logging param when it is an error.
- Add more tests.

### Special notes for reviews

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -131,7 +131,7 @@ func toSafeJSONString(s string) []byte {
 	if data, err := json.Marshal(s); err == nil {
 		return data
 	} else {
-		return []byte(`"JSON Marshal Error"`)
+		return []byte("\"" + s + "\"")
 	}
 }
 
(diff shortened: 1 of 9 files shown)


logging/logging.go
```diff
@@ -126,6 +126,15 @@ func caller(depth int) (file string, line int) {
 	return
 }
 
+// toSafeJSONString converts to valid JSON string, as the original string may contain '\\', '\n', '\r', '\t' and so on.
+func toSafeJSONString(s string) []byte {
+	if data, err := json.Marshal(s); err == nil {
+		return data
+	} else {
+		return []byte(`"JSON Marshal Error"`)
```
louyuting, line 134: Maybe we just return `"` + s + `"`  when occurs error?

logging/logging.go
```diff
@@ -189,23 +197,27 @@ func AssembleMsg(depth int, logLevel, msg string, err error, keysAndValues ...in
 				kStr = fmt.Sprintf("%+v", k)
 			}
 			sb.WriteByte(',')
-			sb.WriteByte('"')
-			sb.WriteString(kStr)
-			sb.WriteByte('"')
+			data := toSafeJSONString(kStr)
+			sb.Write(data)
 			sb.WriteByte(':')
 			vStr, vIsStr := v.(string)
```
louyuting, line 203: Maybe we could refine code here:
```go
                   switch v.(type) {
			case string:
				data := toSafeJSONString(v.(string))
				sb.Write(data)
			case error:
				data := toSafeJSONString(v.(error).Error())
				sb.Write(data)
			default:
				if vbs, err := json.Marshal(v); err != nil {
					s := fmt.Sprintf("%+v", v)
					data := toSafeJSONString(s)
					sb.Write(data)
				} else {
					sb.Write(vbs)
				}
			}
```

luckyxiaoqiang: Good advice. Addressed.

logging/logging.go
luckyxiaoqiang: Agree. Addressed.

louyuting (approved):
nice work and looking forward more

PR #312, opened 2020-10-28
not supported retrieve and update windows system stat

sanxun0325:
Temporarily not supported retrieve and update windows system stat，

### Describe how you did it
Determine the system type, and if it is Windows, do not count load

### Describe how to verify it
Run the example and observe the log

### Special notes for reviews


louyuting (approved):
Nice work

PR #315, opened 2020-11-02
fix bug in circuit_breaker.go about SlowRequestRatio

Jayice-zjw:
when slowRequestRatio was set to 1.0d, requests will **never** be blocked. Because the currentRatio will never be bigger than 100%

### Describe how you did it
Special treatment for case when slowRequestRatio was set to 1.0d

### Special notes for reviews

### DIFF
--- a/core/circuitbreaker/circuit_breaker.go
+++ b/core/circuitbreaker/circuit_breaker.go
@@ -290,7 +290,7 @@ func (b *slowRtCircuitBreaker) OnRequestComplete(rt uint64, err error) {
 		return
 	}
 
-	if slowRatio > b.maxSlowRequestRatio {
+	if slowRatio > b.maxSlowRequestRatio || util.Float64Equals(slowRatio, b.maxSlowRequestRatio) {
 		curStatus = b.CurrentState()
 		switch curStatus {
 		case Closed:
@@ -468,7 +468,7 @@ func (b *errorRatioCircuitBreaker) OnRequestComplete(rt uint64, err error) {
 	if totalCount < b.minRequestAmount {
 		return
 	}
-	if errorRatio >= b.errorRatioThreshold {
+	if errorRatio > b.errorRatioThreshold || util.Float64Equals(errorRatio, b.errorRatioThreshold) {
 		curStatus = b.CurrentState()
 		switch curStatus {
 		case Closed:
--- a/core/circuitbreaker/circuit_breaker.go
+++ b/core/circuitbreaker/circuit_breaker.go
@@ -468,7 +468,7 @@ func (b *errorRatioCircuitBreaker) OnRequestComplete(rt uint64, err error) {
 	if totalCount < b.minRequestAmount {
 		return
 	}
-	if errorRatio > b.errorRatioThreshold {
+	if errorRatio >= b.errorRatioThreshold {
 		curStatus = b.CurrentState()
 		switch curStatus {
 		case Closed:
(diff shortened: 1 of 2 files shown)


sanxun0325 (commented):
Maybe the same problem exists in errorRatioCircuitBreaker

Jayice-zjw:
> Maybe the same problem exists in errorRatioCircuitBreaker

Exactly. Thanks for your tip

louyuting:
The Rule.Threshold means the max value that circuit breaker allowed. If the statistic ratio is great than Threshold, the circuit breaker will open.

From my perspective, setting Rule.Threshold as 1.0 doesn't make sense. May be we should check (Threshold==1.0) in https://github.com/alibaba/sentinel-golang/blob/51b28fdb3acd42bf0f47ac1ac00af2be237f2f55/core/circuitbreaker/rule_manager.go#L347

Jayice-zjw:
> The Rule.Threshold means the max value that circuit breaker allowed. If the statistic ratio is great than Threshold, the circuit breaker will open.
> 
> From my perspective, setting Rule.Threshold as 1.0 doesn't make sense. May be we should check (Threshold==1.0) in
> 
> https://github.com/alibaba/sentinel-golang/blob/51b28fdb3acd42bf0f47ac1ac00af2be237f2f55/core/circuitbreaker/rule_manager.go#L347
You make a good point! But I think it would be better to support this, in case this feature is needed in some scenarios? What do you think?

core/circuitbreaker/circuit_breaker.go
```diff
@@ -290,7 +290,8 @@ func (b *slowRtCircuitBreaker) OnRequestComplete(rt uint64, err error) {
 		return
 	}
 
-	if slowRatio > b.maxSlowRequestRatio {
+	// Special case when maxSlowRequestRatio is set to 1.0 should be handled
```
louyuting, line 293: Maybe we need to change checking logic.
Using "slowRatio >= b.maxSlowRequestRatio"、”errorRatio >= b.errorRatioThreshold“ and ”errorCount >= b.errorCountThreshold“ 

It's easier to understand.

louyuting:
> > The Rule.Threshold means the max value that circuit breaker allowed. If the statistic ratio is great than Threshold, the circuit breaker will open.
> > From my perspective, setting Rule.Threshold as 1.0 doesn't make sense. May be we should check (Threshold==1.0) in
> > https://github.com/alibaba/sentinel-golang/blob/51b28fdb3acd42bf0f47ac1ac00af2be237f2f55/core/circuitbreaker/rule_manager.go#L347
> > 
> > You make a good point! But I think it would be better to support this, in case this feature is needed in some scenarios? What do you think?

Maybe we need improve the checking logic here. Refer to my comments

louyuting:
Maybe  you do a wrong update. Below is my comments:
>Maybe we need to change checking logic.
Using "slowRatio >= b.maxSlowRequestRatio"、”errorRatio >= b.errorRatioThreshold“ and ”errorCount >= b.errorCountThreshold“
It's easier to understand.

Jayice-zjw:
> Maybe you do a wrong update. Below is my comments:
> 
> > Maybe we need to change checking logic.
> > Using "slowRatio >= b.maxSlowRequestRatio"、”errorRatio >= b.errorRatioThreshold“ and ”errorCount >= b.errorCountThreshold“
> > It's easier to understand.

Modified.Thanks for suggestions

louyuting (approved):
Nice work and looking forward more~

Jayice-zjw:
> Nice work and looking forward more~

appreciated for your help

PR #316, opened 2020-11-03
Remove redundant type conversion.

liqiangz:
Remove redundant type conversion.

### Special notes for reviews


louyuting (approved):
Nice work

Issue #317, opened 2020-11-05
Make Slot implementation with Order semantic


louyuting:
## Issue Description

Type: *feature request*

luckyxiaoqiang:
Please assign to me.

louyuting:
refer to PR #318

PR #318, opened 2020-11-08
Support slot order

luckyxiaoqiang:
Support slot order. 

Slots in a slot chain are sorted by order value of slots in increasing order.  Default slot chain has a fixed default order and have no relation to the add sequence. User extended slot can set the order to make it be in front of or at the back of a sentinel build-in slot easily. 

#317 

### Special notes for reviews


core/base/slot_chain.go
```diff
@@ -125,15 +129,16 @@ func (sc *SlotChain) RangeStatPrepareSlot(f func(slot StatPrepareSlot)) {
...
 
-func (sc *SlotChain) AddStatPrepareSlotFirst(s StatPrepareSlot) {
-	ns := make([]StatPrepareSlot, 0, len(sc.statPres)+1)
-	// add to first
-	ns = append(ns, s)
-	sc.statPres = append(ns, sc.statPres...)
-}
-
-func (sc *SlotChain) AddStatPrepareSlotLast(s StatPrepareSlot) {
+func (sc *SlotChain) InsertStatPrepareSlotByOrder(s StatPrepareSlot) {
```
louyuting, line 132: Maybe suffix “ByOrder” is not needed?
such as "AddStatPrepareSlot()" is ok?

core/base/slot_chain.go
```diff
@@ -125,15 +129,16 @@ func (sc *SlotChain) RangeStatPrepareSlot(f func(slot StatPrepareSlot)) {
...
 	sc.statPres = append(sc.statPres, s)
+	i := len(sc.statPres) - 1
```
louyuting, line 134: Can the standard pkg "sort" support sorting logic?
such as ‘sort.Slice()’？

luckyxiaoqiang: Refined.

core/base/slot_chain.go
luckyxiaoqiang: Refined.

core/base/slot_chain.go
```diff
@@ -157,14 +158,11 @@ func (sc *SlotChain) RangeRuleCheckSlot(f func(slot RuleCheckSlot)) {
...
-func (sc *SlotChain) AddRuleCheckSlotFirst(s RuleCheckSlot) {
-	ns := make([]RuleCheckSlot, 0, len(sc.ruleChecks)+1)
-	ns = append(ns, s)
-	sc.ruleChecks = append(ns, sc.ruleChecks...)
-}
-
-func (sc *SlotChain) AddRuleCheckSlotLast(s RuleCheckSlot) {
+func (sc *SlotChain) AddRuleCheckSlot(s RuleCheckSlot) {
 	sc.ruleChecks = append(sc.ruleChecks, s)
+	sort.Slice(sc.ruleChecks, func(i, j int) bool {
```
louyuting, line 163: Maybe we need unit test to cover rank result.

core/base/slot_chain.go
```diff
@@ -1,22 +1,27 @@
...
+	// Order returns the order value of slot.
+	// Slots will be sorted according to order value in increasing order.
+	Order() uint32
 }
 
 // StatPrepareSlot is responsible for some preparation before statistic
 // For example: init structure and so on
 type StatPrepareSlot interface {
-	NamedSlot
+	BaseSlot
```
louyuting, line 24: Maybe we need some comments to describe the order logic.
StatPrepareSlot、RuleCheckSlot、StatSlot is the independent bucket, the ranking logic is independent.

core/base/slot_chain.go
```diff
@@ -28,7 +33,7 @@ type StatPrepareSlot interface {
 // RuleCheckSlot is rule based checking strategy
 // All checking rule must implement this interface.
 type RuleCheckSlot interface {
-	NamedSlot
+	BaseSlot
```
louyuting, line 36: Same as upon

core/base/slot_chain.go
```diff
@@ -39,7 +44,7 @@ type RuleCheckSlot interface {
 // StatSlot is responsible for counting all custom biz metrics.
 // StatSlot would not handle any panic, and pass up all panic to slot chain
 type StatSlot interface {
-	NamedSlot
+	BaseSlot
```
louyuting, line 47: Same with upon

louyuting:
Some suggestion about default order of embedded  slot.
1. In each bucket(PrepareStatSlot Bucket、CheckSlot Bucket、StatSlot Bucket), The default order should from local to external。Such as CheckSlot, it should be system、flow、isolation、hotspot、circuitbreaker;
2. The StatSlot order should keep consistent with CheckSlot bucket。 Such as: flowStat、hotspotStat、circuitbreakerStat.

core/base/slot_chain.go
luckyxiaoqiang: There are already related unit tests like `TestSlotChain_AddStatPrepareSlot`. Is is sufficient?

louyuting: Okay for me.

luckyxiaoqiang:
> Some suggestion about default order of embedded slot.
> 
> 1. In each bucket(PrepareStatSlot Bucket、CheckSlot Bucket、StatSlot Bucket), The default order should from local to external。Such as CheckSlot, it should be system、flow、isolation、hotspot、circuitbreaker;
> 2. The StatSlot order should keep consistent with CheckSlot bucket。 Such as: flowStat、hotspotStat、circuitbreakerStat.

Agree. Refined.

louyuting:
@liqiangz @sanxun0325  Could you please take a look on this PR ?

core/circuitbreaker/slot.go
```diff
@@ -5,7 +5,8 @@ import (
 )
 
 const (
-	RuleCheckSlotName = "sentinel-core-circuit-breaker-rule-check-slot"
+	RuleCheckSlotName  = "sentinel-core-circuit-breaker-rule-check-slot"
+	RuleCheckSlotOrder = 5000
```
sanxun0325, line 9: It is better to define the order of slots in a uniform location, which may seem more intuitive。

core/stat/stat_prepare_slot.go
```diff
@@ -5,7 +5,8 @@ import (
 )
 
 const (
-	PrepareSlotName = "sentinel-core-stat-resource-node-prepare-slot"
+	PrepareSlotName  = "sentinel-core-stat-resource-node-prepare-slot"
+	PrepareSlotOrder = 1000
```
sanxun0325, line 9: Does the definition of a slot order, starting at 1000, have any particular meaning?

core/circuitbreaker/slot.go
luckyxiaoqiang: My first version is like what you recommended, current version is the result of discussion with @louyuting .

core/stat/stat_prepare_slot.go
luckyxiaoqiang: Leave a gap for the user to add extended slots before it.

sanxun0325: > Leave a gap for the user to add extended slots before it.

Yes,The main thing I'm trying to say is the order of slots 1000, why isn't it 2000, or 10000.There should be a specification for it.  I don't know if there are any other better ideas for defining the slot order that Sentinel provides by default。

louyuting (approved):

sczyh30 (approved):
LGTM

PR #320, opened 2020-11-09
Refine logging

luckyxiaoqiang:
Refine logging. If a string is already in json format, do not marshal again.

### Special notes for reviews


louyuting (approved):
Nice work

PR #322, opened 2020-11-11
Loading isolation rule supports cache

sanxun0325:
Describe what this PR does / why we need it

Does this pull request fix one issue?
Fix:#111

Describe how you did it

Describe how to verify it
ut

Special notes for reviews

### DIFF
--- a/core/isolation/rule_manager.go
+++ b/core/isolation/rule_manager.go
@@ -1,24 +1,37 @@
 package isolation
 
 import (
+	"reflect"
 	"sync"
 
 	"github.com/alibaba/sentinel-golang/core/misc"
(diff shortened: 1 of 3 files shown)


core/isolation/rule_manager.go
```diff
@@ -10,15 +11,27 @@ import (
 )
 
 var (
-	ruleMap = make(map[string][]*Rule)
-	rwMux   = &sync.RWMutex{}
+	ruleMap      = make(map[string][]*Rule)
+	rwMux        = &sync.RWMutex{}
+	currentRules = make([]*Rule, 0, 4)
```
louyuting, line 16: initial size is unnecessary.

louyuting (approved):

PR #323, opened 2020-11-11
Add isolation rules handler

sanxun0325:
Add isolation rules handler

### Describe how you did it
IsolationRuleJsonArrayParser provide JSON  as the default serialization for list of isolation.Rule
IsolationRulesUpdater load the newest []isolation.Rule to downstream system component.

### Describe how to verify it
ut

### Special notes for reviews

### DIFF
--- a/ext/datasource/helper.go
+++ b/ext/datasource/helper.go
@@ -4,11 +4,10 @@ import (
 	"encoding/json"
 	"fmt"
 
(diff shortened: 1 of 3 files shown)


louyuting (approved):
Nice work and looking forward more.

louyuting (approved):
Nice work and looking forward more

PR #342, opened 2020-12-06
Rule Manager: Support update rule hook fix #275

### LINKED ISSUE #275: Rule Manager: Abstract general hook for updating rule
## Issue Description

Type: *feature request*

Currently, the logic of LoadRules is very heavy, there is some adjoint logic, such as: register resource slot chain、log rules....
Maybe we should abstract internal general hook for updating rules.

Limitation: internal changes and this hook is invisible for user.

CrazyHZM:
support update rule hook

Fixes #275

### Special notes for reviews

### DIFF
--- a/core/circuitbreaker/rule_manager.go
+++ b/core/circuitbreaker/rule_manager.go
@@ -16,11 +16,11 @@ type CircuitBreakerGenFunc func(r *Rule, reuseStat interface{}) (CircuitBreaker,
 var (
 	cbGenFuncMap = make(map[Strategy]CircuitBreakerGenFunc, 4)
 
-	breakerRules = make(map[string][]*Rule)
-	breakers     = make(map[string][]CircuitBreaker)
-	updateMux    = &sync.RWMutex{}
-	currentRules = make([]*Rule, 0)
-
+	BreakerRules         = make(map[string][]*Rule)
+	Breakers             = make(map[string][]CircuitBreaker)
+	UpdateMux            = &sync.RWMutex{}
+	CurrentRules         = make([]*Rule, 0)
+	ruleUpdateHandler    = DefaultRuleUpdateHandler
 	stateChangeListeners = make([]StateChangeListener, 0)
 )
 
@@ -76,9 +76,9 @@ func init() {
 // GetRulesOfResource need to compete circuit breaker module's global lock and the high performance losses of copy,
 // 		reduce or do not call GetRulesOfResource frequently if possible
 func GetRulesOfResource(resource string) []Rule {
-	updateMux.RLock()
-	resRules, ok := breakerRules[resource]
-	updateMux.RUnlock()
+	UpdateMux.RLock()
+	resRules, ok := BreakerRules[resource]
+	UpdateMux.RUnlock()
 	if !ok {
 		return nil
 	}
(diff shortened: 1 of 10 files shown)


sanxun0325 (commented):
Can you describe the design of this PR?

core/circuitbreaker/rule_manager.go
```diff
@@ -16,11 +16,11 @@ type CircuitBreakerGenFunc func(r *Rule, reuseStat interface{}) (CircuitBreaker,
 var (
 	cbGenFuncMap = make(map[Strategy]CircuitBreakerGenFunc, 4)
 
-	breakerRules = make(map[string][]*Rule)
-	breakers     = make(map[string][]CircuitBreaker)
-	updateMux    = &sync.RWMutex{}
-	currentRules = make([]*Rule, 0)
-
+	BreakerRules         = make(map[string][]*Rule)
```
sanxun0325, line 19: Why change the scope of a variable? It's not safe

CrazyHZM:
> Can you describe the design of this PR?

There are two scenarios:
1.Users hope to have some processing operations before and after updating the rules. For example:
```go
func LoadRule()  {
   beforeUpdateRule()
   updateRule()
   afterUpdateRule()
}
```
2.Users want to be able to directly rewrite the update operation. For example:
rewriite IsValidRule(rule *Rule)
rewriite logRuleUpdate() ...

The first scenario can be implemented directly through LoadRule, but the second scenario needs to  provide a way to manipulate variables.

core/circuitbreaker/rule_manager.go
CrazyHZM: I agree with you, but can modify a variable through function, I will modify it.

liqiangz:
@CrazyHZM
I think we should not modify the original functions and expose variable.  We should only allow expansion of the original.
It is sufficient to provide the following method.
```go
func registerRuleUpdateHandler(handler func(onRuleUpdate func(rules []*Rule) (err error))) (err error))
```

CrazyHZM:
> I think we should not modify the original functions and expose variable. We should only allow expansion of the original.
> I have a proposal :
> 
> It is sufficient to provide the following method.
> 
> ```go
> type Handler func(onRuleUpdate func(rules []*Rule) (err error), rules []*Rule)
> 
> func RegisterRuleUpdateHandler(handler Handler)
> 
> RegisterRuleUpdateHandler(func(onRuleUpdate func(rules []*Rule) (err error), rules []*Rule){
> 		beforeUpdateRule()
> 		if IsValidRule() {
> 			onRuleUpdate(rules)
> 		}
> 		afterUpdateRule()
> 	})
> ```
> 
> What is your opinion about it? @CrazyHZM

This can also meet the needs without destroying the original design. I think it is a good idea. I have modified it.

core/hotspot/rule_manager.go
```diff
@@ -80,7 +83,7 @@ func LoadRules(rules []*Rule) (bool, error) {
 		return false, nil
```
liqiangz, line 83: Maybe we can move isEqual judge logic to onRuleUpdate function. Otherwise, there may be some equal cases without going to our registered handler.

core/isolation/rule_manager.go
```diff
@@ -158,3 +161,11 @@ func IsValid(r *Rule) error {
 	}
 	return nil
 }
+
+func RegisterRuleUpdateHandler(handler RuleUpdateHandler) {
+	ruleUpdateHandler = handler
```
liqiangz, line 166: This method should be thread safe

core/circuitbreaker/rule_manager.go
```diff
@@ -13,15 +13,17 @@ import (
...
 
-	breakerRules = make(map[string][]*Rule)
-	breakers     = make(map[string][]CircuitBreaker)
-	updateMux    = &sync.RWMutex{}
-	currentRules = make([]*Rule, 0)
-
+	breakerRules         = make(map[string][]*Rule)
+	breakers             = make(map[string][]CircuitBreaker)
+	updateMux            = &sync.RWMutex{}
+	currentRules         = make([]*Rule, 0)
```
luckyxiaoqiang, line 24: Why change the format?

louyuting:
https://github.com/alibaba/sentinel-golang/issues/275#issuecomment-764628023

louyuting:
refer to https://github.com/alibaba/sentinel-golang/issues/275#issuecomment-764628023
close it.
If you have any discussion, please reach out to me.

PR #350, opened 2020-12-13
Support mock time and ticker

luckyxiaoqiang:
Support mock time and ticker. We can use mock time and ticker to do unit test, which can decouple from system real time. 

### Special notes for reviews

### DIFF
--- a/core/flow/slot.go
+++ b/core/flow/slot.go
@@ -15,11 +15,10 @@
 package flow
 
 import (
-	"time"
-
 	"github.com/alibaba/sentinel-golang/core/base"
 	"github.com/alibaba/sentinel-golang/core/stat"
 	"github.com/alibaba/sentinel-golang/logging"
+	"github.com/alibaba/sentinel-golang/util"
 	"github.com/pkg/errors"
 )
 
@@ -65,7 +64,7 @@ func (s *Slot) Check(ctx *base.EntryContext) *base.TokenResult {
 		if r.Status() == base.ResultStatusShouldWait {
 			if nanosToWait := r.NanosToWait(); nanosToWait > 0 {
 				// Handle waiting action.
-				time.Sleep(nanosToWait)
+				util.Sleep(nanosToWait)
 			}
 			continue
 		}
--- a/util/time.go
+++ b/util/time.go
@@ -62,43 +62,78 @@ type Clock interface {
 	CurrentTimeNano() uint64
 }
 
+// clockWrapper is used for atomic operation.
+type clockWrapper struct {
+	clock Clock
+}
+
+// RealClock wraps some APIs of time package.
+type RealClock struct{}
+
+func NewRealClock() *RealClock {
+	return &RealClock{}
+}
+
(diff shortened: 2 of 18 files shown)


core/flow/slot.go
```diff
@@ -44,6 +43,8 @@ func (s *Slot) Order() uint32 {
 }
 
 func (s *Slot) Check(ctx *base.EntryContext) *base.TokenResult {
+	util.SetClock(util.NewMockClock())
```
sczyh30, line 46: Is it needed here?

luckyxiaoqiang: > Is it needed here?

That's a mistake.

util/time.go
```diff
@@ -24,6 +26,261 @@ const (
...
+
+	_ Ticker = &RealTicker{}
+	_ Ticker = &MockTicker{}
+
+	_ TickerCreator = &RealTickerCreator{}
+	_ TickerCreator = &MockTickerCreator{}
+)
+
+var (
+	realClock    *RealClock
```
louyuting, line 41: Maybe this field is unnecessary?

util/time.go
```diff
@@ -24,6 +26,261 @@ const (
...
+	currentClock *atomic.Value
+
+	realTickerCreator    *RealTickerCreator
```
louyuting, line 44: Maybe this field is unnecessary?

util/time.go
```diff
@@ -36,15 +293,21 @@ func FormatDate(tsMillis uint64) string {
...
 }
 
 // Returns the current Unix timestamp in nanoseconds.
 func CurrentTimeNano() uint64 {
-	return uint64(time.Now().UnixNano())
+	return CurrentClock().CurrentTimeNano()
+}
+
+// Now returns the current local time.
+func Now() time.Time {
```
louyuting, line 305: Maybe we need a benchmark to test the performance diff between `Now()` and `CurrentTimeMillsWithTicker`

util/time.go
luckyxiaoqiang: Now it is used to do a safe fallback like in `CurrentClock`.

util/time.go
luckyxiaoqiang: Now it is used to do a safe fallback like in `CurrentTickerCreator`.

util/time.go
luckyxiaoqiang: Refined.

util/time.go
luckyxiaoqiang: Added. 
```
BenchmarkCurrentTimeInMs/CurrentTimeMillis-4            230781794                5.26 ns/op
BenchmarkCurrentTimeInMs/CurrentTimeMillisDirect-4      519137175                2.32 ns/op
BenchmarkNow/util.Now-4                                 14527578                83.3 ns/op             
BenchmarkNow/time.Now-4                                 14849454                77.0 ns/op             
```

louyuting (approved):
LGTM

sczyh30 (approved):
LGTM

PR #358, opened 2020-12-20
Isolate system metric statistic to independent pkg core/system_metric

louyuting:
Isolate system metric statistic to independent pkg core/system_metric

### Special notes for reviews


sanxun0325 (dismissed):
LGTM

luckyxiaoqiang (approved):
LGTM

PR #360, opened 2020-12-22
热点规则增加ParamKey，方便快速从大量的键值对中指定需要热点限流的值 

cafra:
热点参数规则中，增加ParamKey配合context attachments map ,方便快速从大量的键值对中指定需要热点限流的值. (alibaba#355)

方便热点限流规则中，指定热点参数。
比如把系统参数、header中标定参数、业务参数，数据很多，不同业务需要的参数也不相同，加入把全部数据放入ctx.attachments 这样在规则中指定特点的key，很方便。

本次pr 已经增加ParamKey, 并且修改了对应的单元测试和example.

### Describe how you did it
规则增加ParamKey；
trafficShaper 也对应实现extractArg 接口，可以与已有的paramIndex 一起方便指定热点参数。

### Describe how to verify it
单元测试和example 中均验证过了

### Special notes for reviews
ParamKey 可以考虑变成[]string ，这样使用者除了指定单个参数外，还可以灵活的指定热点的规则。
例如ParamKey=[]string{"uid","merchant_id"} 这样可以让uid和merchant_id 作为联合参数作为热点。

### DIFF
--- a/core/hotspot/concurrency_stat_slot.go
+++ b/core/hotspot/concurrency_stat_slot.go
@@ -15,6 +15,8 @@
 package hotspot
 
 import (
+	"context"
+	"golang.org/x/sync/errgroup"
 	"sync/atomic"
 
 	"github.com/alibaba/sentinel-golang/core/base"
@@ -44,7 +46,9 @@ func (s *ConcurrencyStatSlot) Order() uint32 {
 
 func (c *ConcurrencyStatSlot) OnEntryPassed(ctx *base.EntryContext) {
 	res := ctx.Resource.Name()
+	g, _ := errgroup.WithContext(context.Background())
 	tcs := getTrafficControllersFor(res)
+
 	for _, tc := range tcs {
 		if tc.BoundRule().MetricType != Concurrency {
 			continue
@@ -54,17 +58,23 @@ func (c *ConcurrencyStatSlot) OnEntryPassed(ctx *base.EntryContext) {
 			continue
 		}
 		for _, arg := range args {
-			metric := tc.BoundMetric()
-			concurrencyPtr, existed := metric.ConcurrencyCounter.Get(arg)
-			if !existed || concurrencyPtr == nil {
-				if logging.DebugEnabled() {
-					logging.Debug("[ConcurrencyStatSlot OnEntryPassed] Parameter does not exist in ConcurrencyCounter.", "argument", arg)
+			arg := arg // https://golang.org/doc/faq#closures_and_goroutines
+			g.Go(func() error {
+				metric := tc.BoundMetric()
+				concurrencyPtr, existed := metric.ConcurrencyCounter.Get(arg)
+				if !existed || concurrencyPtr == nil {
+					if logging.DebugEnabled() {
+						logging.Debug("[ConcurrencyStatSlot OnEntryPassed] Parameter does not exist in ConcurrencyCounter.", "argument", arg)
+					}
+					return nil
 				}
-				continue
-			}
-			atomic.AddInt64(concurrencyPtr, 1)
+				atomic.AddInt64(concurrencyPtr, 1)
+				return nil
+			})
 		}
 	}
+	g.Wait()
+	return
 }
 
 func (c *ConcurrencyStatSlot) OnEntryBlocked(ctx *base.EntryContext, blockError *base.BlockError) {
@@ -73,7 +83,9 @@ func (c *ConcurrencyStatSlot) OnEntryBlocked(ctx *base.EntryContext, blockError 
 
 func (c *ConcurrencyStatSlot) OnCompleted(ctx *base.EntryContext) {
 	res := ctx.Resource.Name()
+	g, _ := errgroup.WithContext(context.Background())
 	tcs := getTrafficControllersFor(res)
+
 	for _, tc := range tcs {
 		if tc.BoundRule().MetricType != Concurrency {
 			continue
@@ -83,15 +95,21 @@ func (c *ConcurrencyStatSlot) OnCompleted(ctx *base.EntryContext) {
 			continue
 		}
 		for _, arg := range args {
-			metric := tc.BoundMetric()
-			concurrencyPtr, existed := metric.ConcurrencyCounter.Get(arg)
-			if !existed || concurrencyPtr == nil {
-				if logging.DebugEnabled() {
-					logging.Debug("[ConcurrencyStatSlot OnCompleted] Parameter does not exist in ConcurrencyCounter.", "argument", arg)
+			arg := arg // https://golang.org/doc/faq#closures_and_goroutines
+			g.Go(func() error {
+				metric := tc.BoundMetric()
+				concurrencyPtr, existed := metric.ConcurrencyCounter.Get(arg)
+				if !existed || concurrencyPtr == nil {
+					if logging.DebugEnabled() {
+						logging.Debug("[ConcurrencyStatSlot OnCompleted] Parameter does not exist in ConcurrencyCounter.", "argument", arg)
+					}
+					return nil
 				}
-				continue
-			}
-			atomic.AddInt64(concurrencyPtr, -1)
+				atomic.AddInt64(concurrencyPtr, -1)
+				return nil
+			})
 		}
 	}
+	g.Wait()
+	return
 }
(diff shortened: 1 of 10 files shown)


core/hotspot/concurrency_stat_slot.go
```diff
@@ -44,26 +47,35 @@ func (s *ConcurrencyStatSlot) Order() uint32 {
...
-		metric := tc.BoundMetric()
-		concurrencyPtr, existed := metric.ConcurrencyCounter.Get(arg)
-		if !existed || concurrencyPtr == nil {
-			if logging.DebugEnabled() {
-				logging.Debug("[ConcurrencyStatSlot OnEntryPassed] Parameter does not exist in ConcurrencyCounter.", "argument", arg)
-			}
-			continue
+		for _, arg := range args {
+			arg := arg // https://golang.org/doc/faq#closures_and_goroutines
+			g.Go(func() error {
```
louyuting, line 63: Could you please provide some benchmark data to verify why introduce goroutine group?

core/hotspot/slot.go
```diff
@@ -15,10 +15,13 @@
 package hotspot
 
 import (
+	"context"
 	"time"
 
+	"golang.org/x/sync/errgroup"
+	"gopkg.in/errgo.v2/fmt/errors"
```
louyuting, line 22: Maybe `github.com/pkg/errors`  is enough?

go.mod
```diff
@@ -9,5 +9,7 @@ require (
 	github.com/shirou/gopsutil/v3 v3.20.11
 	github.com/stretchr/testify v1.6.1
 	go.uber.org/multierr v1.5.0
+	golang.org/x/sync v0.0.0-20190423024810-112230192c58
+	gopkg.in/errgo.v2 v2.1.0
```
louyuting, line 13: Maybe `github.com/pkg/errors` is enough?

core/hotspot/concurrency_stat_slot.go
sanxun0325: There is no return of error here, and perhaps the error group is not required.

core/hotspot/traffic_shaping.go
```diff
@@ -155,6 +159,74 @@ func (c *baseTrafficShapingController) BoundParamIndex() int {
...
+	}
+	return args[idx]
+}
+func (c *baseTrafficShapingController) extractAttachmentArgs(ctx *base.EntryContext) interface{} {
+	attachments := ctx.Input.Attachments
+
+	if attachments == nil {
+		if logging.DebugEnabled() {
+			logging.Debug("[paramKey] The attachments of ctx is nil",
+				"args", attachments, "paramIndex", c.paramKey)
```
sanxun0325, line 208: `"paramIndex"` it should be paramKey,   `[paramKey]` it should be [BaseTrafficShapingController extractAttachmentArgs].

louyuting:
Could you please provide some benchmark for involving errgroup?

core/hotspot/concurrency_stat_slot.go
```diff
@@ -72,24 +84,33 @@ func (c *ConcurrencyStatSlot) OnEntryBlocked(ctx *base.EntryContext, blockError
...
 		}
-		metric := tc.BoundMetric()
-		concurrencyPtr, existed := metric.ConcurrencyCounter.Get(arg)
-		if !existed || concurrencyPtr == nil {
-			if logging.DebugEnabled() {
-				logging.Debug("[ConcurrencyStatSlot OnCompleted] Parameter does not exist in ConcurrencyCounter.", "argument", arg)
-			}
-			continue
+		for _, arg := range args {
+			arg := arg // https://golang.org/doc/faq#closures_and_goroutines
```
liqiangz, line 99: Maybe  we can pass the variable as an argument to the closure

core/hotspot/traffic_shaping.go
```diff
@@ -155,6 +159,74 @@ func (c *baseTrafficShapingController) BoundParamIndex() int {
 	return c.paramIndex
 }
 
+// ExtractArgs matches the arg from ctx based on TrafficShapingController
+// return nil if match failed.
+func (c *baseTrafficShapingController) ExtractArgs(ctx *base.EntryContext) []interface{} {
+	if c == nil {
+		return nil
+	}
+	args := make([]interface{}, 0, len(ctx.Input.Args)+len(ctx.Input.Attachments))
```
liqiangz, line 168: The maxLen is 2,  Maybe we can set len=2.

core/hotspot/concurrency_stat_slot.go
cafra: The err group only throws an error when a block occurs, and quickly ends the concurrent check @sanxun0325

core/hotspot/slot.go
cafra: yes, it's enough ,I will replace it

go.mod
cafra: yes

core/hotspot/slot.go
cafra: > Could you please provide some benchmark for involving errgroup?

ok ,i will provide it today

louyuting: Nice~

core/hotspot/traffic_shaping.go
cafra: The length of arg and attach is not determined.
Not all of the attach parameters are used as hot spots.
So the length of arg+attach should be more, but the length is measurable.
Why set the length to 2?

core/hotspot/traffic_shaping.go
cafra: When there is no error, errgroup becomes waitgroup.
The purpose is to count concurrently. Reduce processing time.

paramIndex has been replaced by paramKey.

cafra:
> Could you please provide some benchmark for involving errgroup?

## The factors affecting performance are as follows and the larger the factor data, the better the concurrent performance
1. The number of extraction parameters; 
2. Process time; 
3. Hit rate；

## Mock code
```golang
package main

import (
	"context"
	"time"

	"golang.org/x/sync/errgroup"
)

func main() {

}

func do(duration time.Duration) {
	time.Sleep(duration)
}

func forRange(n int, duration time.Duration) {
	for i := 0; i < n; i++ {
		do(duration)
	}
}

func ConRange(n int, duration time.Duration) {
	var g *errgroup.Group
	if n == 0 {
		return
	}
	g, _ = errgroup.WithContext(context.Background())

	for i := 0; i < n; i++ {
		g.Go(func() error {
			do(duration)
			return nil
		})
	}
	if err := g.Wait(); err != nil {
		return
	}
}

```

## Bench Code

```golang
package main

import (
	"fmt"
	"testing"
	"time"
)

func BenchmarkDo(b *testing.B) {
	b.ReportAllocs()
	b.ResetTimer()

	for _, duration := range []int{1, 5, 10} {
		for _, c := range []int{0, 1, 2, 3, 4, 8, 12} {
			dt := time.Duration(duration) * time.Microsecond
			k1 := fmt.Sprintf("for(%v) duration(%v micro)", c, duration)
			b.Run(k1, func(b *testing.B) {
				for i := 0; i < b.N; i++ {
					forRange(c, dt)
				}
			})
			k2 := fmt.Sprintf("con(%v) duration(%v micro)", c, duration)
			b.Run(k2, func(b *testing.B) {
				for i := 0; i < b.N; i++ {
					ConRange(c, dt)
				}
			})
		}
	}
}

```

## Result Data
image

## Conclusion
When more than 2 parameters are extracted, concurrent processing is used.
Otherwise, normal loop processing.

core/hotspot/traffic_shaping.go
liqiangz: Currently args only append c.extractArgs(ctx) and c.extractAttachmentArgs(ctx), the maxLen is 2.  len(ctx.Input.Args)+len(ctx.Input.Attachments) may be too big?

cafra: Remove the slice, paramIndex and paramKey are mutually exclusive to avoid performance loss.

is it ok?

core/hotspot/concurrency_stat_slot.go
sanxun0325: Yes, that's right.but the current groutine doesn't throw an error

core/hotspot/traffic_shaping.go
liqiangz: ok

louyuting:
Duplicated : https://github.com/alibaba/sentinel-golang/pull/376

PR #373, opened 2021-01-06
Support resource level rule replacement in hotspot module

luckyxiaoqiang:
Support resource level rule replacement in hotspot module.

### Special notes for reviews

### DIFF
--- a/core/hotspot/rule_manager.go
+++ b/core/hotspot/rule_manager.go
@@ -82,10 +82,10 @@ func getTrafficControllersFor(res string) []TrafficShapingController {
 	return tcMap[res]
 }
 
-// LoadRules replaces old rules with the given hotspot parameter flow control rules. Return value:
-//
-// bool: indicates whether the internal map has been changed;
-// error: indicates whether occurs the error.
+// LoadRules replaces all old hotspot param flow rules with the given rules.
+// Return value:
+//   bool: indicates whether the internal map has been changed;
+//   error: indicates whether occurs the error.
 func LoadRules(rules []*Rule) (bool, error) {
 	resRulesMap := make(map[string][]*Rule, 16)
 	for _, rule := range rules {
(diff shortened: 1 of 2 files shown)


core/hotspot/rule_manager.go
```diff
@@ -137,7 +146,13 @@ func ClearRules() error {
 	return err
 }
 
-func onRuleUpdate(rules []*Rule) (err error) {
+// ClearRulesOfResource clears resource level parameter flow rules.
```
liqiangz, line 149: flow rules -> hotspot rules

core/hotspot/rule_manager.go
```diff
@@ -237,14 +210,98 @@ func onRuleUpdate(rules []*Rule) (err error) {
 
 	tcMux.Lock()
 	tcMap = m
-	currentRules = rules
+	currentRules = rawResRulesMap
```
liqiangz, line 213: We can do this without lock.

core/hotspot/rule_manager.go
```diff
@@ -237,14 +210,98 @@ func onRuleUpdate(rules []*Rule) (err error) {
...
+
+	tcMux.Lock()
+	if len(newResTcs) == 0 {
+		delete(tcMap, res)
+	} else {
+		tcMap[res] = newResTcs
+	}
+	tcMux.Unlock()
+	currentRules[res] = rawResRules
+	logging.Debug("[HotSpot onResourceRuleUpdate] Time statistic(ns) for updating flow rule", "timeCost", util.CurrentTimeNano()-start)
```
liqiangz, line 267: flow rules -> hotspot rules

core/hotspot/rule_manager.go
luckyxiaoqiang: `parameter flow rules` is the history name, it ` means `hotspot rules`.

core/hotspot/rule_manager.go
luckyxiaoqiang: Updated.

core/hotspot/rule_manager.go
luckyxiaoqiang: Updated.

sanxun0325 (approved):
LGTM

liqiangz (approved):
LGTM

louyuting (approved):

PR #375, opened 2021-01-06
upgrade gopsutil v3.20.12

sanxun0325:
upgrade gopsutil v3.20.12->v3.20.12
v3.20.12 support cpu load on windows

### Special notes for reviews


louyuting (approved):
LGTM

PR #376, opened 2021-01-10
feat: hotspot rule add paramKey;

cafra:
In the hotspot parameter rule, ParamKey is added to cooperate with the context attachments map to quickly specify the value that requires hotspot current limit from a large number of key-value pairs.
(alibaba#360)

It is convenient to specify hot spot parameters in hot spot current limiting rules. For example, there are a lot of data for system parameters, calibration parameters in header, and business parameters, and the parameters required by different businesses are different. It is very convenient to add all the data into ctx.attachments and specify the key in the rules.

This time pr has added ParamKey, and modified the corresponding unit test and example.

### Describe how you did it
ParamKey is added to the rule; trafficShaper also implements the extractArg interface correspondingly, which can be used with the existing paramIndex to easily specify hotspot parameters.

### Describe how to verify it
Verified in unit test and example

### Special notes for reviews
ParamKey can be considered as []string, so that users can specify hotspot rules flexibly in addition to specifying a single parameter. 
For example, ParamKey=[]string{"uid","merchant_id"} This allows uid and merchant_id to be used as joint parameters as hot spots.


core/hotspot/rule_manager.go
```diff
@@ -431,6 +431,9 @@ func IsValidRule(rule *Rule) error {
 	if rule.MetricType == QPS && rule.DurationInSec <= 0 {
 		return errors.New("invalid duration")
 	}
+	if rule.ParamIndex > 0 && rule.ParamKey != "" {
```
louyuting, line 434: it should be `rule.ParamIndex >= 0`

cafra: In order to avoid the problem of paramInde default value 0, 
adjust the order of extracting parameters, 
that is, give priority to paramKey. 
The judgment conditions here will not be adjusted temporarily.

core/hotspot/rule.go
```diff
@@ -80,6 +80,10 @@ type Rule struct {
 	// if ParamIndex is great than or equals to zero, ParamIndex means the <ParamIndex>-th parameter
 	// if ParamIndex is the negative, ParamIndex means the reversed <ParamIndex>-th parameter
 	ParamIndex int `json:"paramIndex"`
+	// ParamKey is the key in context attachments map.
+	// ParamKey can be used as a supplement to ParamIndex to facilitate rules to quickly obtain parameter from a large number of parameters
+	// Mutually exclusive with ParamIndex
```
louyuting, line 85: Maybe we need some comments to illustrate the priority of  ParamIndex and ParamKey

cafra: ok

core/hotspot/rule.go
```diff
@@ -80,6 +80,10 @@ type Rule struct {
...
+	ParamKey string `json:"paramKey"`
```
louyuting, line 86: the function `func (r *Rule) Equals(newRule *Rule) bool` need check this field：`ParamKey `

louyuting: Otherwise, Maybe we need add  another unit test for `func (r *Rule) Equals(newRule *Rule) bool`

core/hotspot/rule.go
```diff
@@ -100,8 +104,8 @@ func (r *Rule) String() string {
 	b, err := json.Marshal(r)
 	if err != nil {
 		// Return the fallback string
-		return fmt.Sprintf("{Id:%s, Resource:%s, MetricType:%+v, ControlBehavior:%+v, ParamIndex:%d, Threshold:%d, MaxQueueingTimeMs:%d, BurstCount:%d, DurationInSec:%d, ParamsMaxCapacity:%d, SpecificItems:%+v}",
-			r.ID, r.Resource, r.MetricType, r.ControlBehavior, r.ParamIndex, r.Threshold, r.MaxQueueingTimeMs, r.BurstCount, r.DurationInSec, r.ParamsMaxCapacity, r.SpecificItems)
+		return fmt.Sprintf("{Id:%s, Resource:%s, MetricType:%+v, ControlBehavior:%+v, ParamIndex:%d, ParamKey:%v, Threshold:%d, MaxQueueingTimeMs:%d, BurstCount:%d, DurationInSec:%d, ParamsMaxCapacity:%d, SpecificItems:%+v}",
```
liqiangz, line 107: Maybe  ParamKey:%v  =>   ParamKey:%s.

cafra: "%s" is better, already updated

louyuting (approved):

sanxun0325 (approved):
LGTM

PR #377, opened 2021-01-12
feat: add NewCustomizedBaseStatNode to create the BaseStatNode with specified param

ansiz:
feat: add NewCustomizedBaseStatNode to create the BaseStatNode with specified param

We need to create the BaseStatNode of a specific size without updating global config.

### Special notes for reviews

### DIFF
--- a/core/stat/base_node.go
+++ b/core/stat/base_node.go
@@ -44,6 +44,23 @@ func NewBaseStatNode(sampleCount uint32, intervalInMs uint32) *BaseStatNode {
 	}
 }
 
+// NewCustomizedBaseStatNode creates the BaseStatNode with specified sampleCount
+// and intervalInMs.
+func NewCustomizedBaseStatNode(sampleCount uint32, intervalInMs uint32) (*BaseStatNode, error) {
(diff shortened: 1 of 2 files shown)


core/stat/base_node.go
```diff
@@ -44,6 +44,23 @@ func NewBaseStatNode(sampleCount uint32, intervalInMs uint32) *BaseStatNode {
 	}
 }
 
+// NewCustomizedBaseStatNode creates the BaseStatNode with specified sampleCount
+// and intervalInMs.
+func NewCustomizedBaseStatNode(sampleCount uint32, intervalInMs uint32) (*BaseStatNode, error) {
```
sanxun0325, line 49: NewCustomizedBaseStatNode  it is not called and can be implemented in your own code

ansiz: But the fields of BaseStatNode are unexported, how do I implement it with my own code? 😢

sanxun0325: I didn't notice this branch, you can do that

sczyh30 (approved):

PR #382, opened 2021-01-16
Support prometheus metrics

luckyxiaoqiang:
Support prometheus metrics.

### Special notes for reviews


exporter/metric/exporter.go
```diff
@@ -0,0 +1,146 @@
...
+	exporter Exporter
+
+	host      string
+	app       string
+	pid       string
+	namespace string
+)
+
+func init() {
+	exporter = newPrometheusExporter()
```
louyuting, line 36: We might need to provide func to reset Exporter

exporter/metric/exporter.go
```diff
@@ -0,0 +1,146 @@
...
+
+	host, _ = os.Hostname()
+	app = config.AppName()
+	pid = strconv.Itoa(os.Getpid())
+	namespace = "sentinel"
```
louyuting, line 41: maybe `sentinel-go`?

exporter/metric/exporter.go
```diff
@@ -0,0 +1,146 @@
...
+
+func (e *prometheusExporter) NewHistogram(name, desc string, buckets []float64, labelNames []string) Histogram {
+	return prometheus.NewHistogram(name, namespace, desc, buckets, labelNames, newConstLabels())
+}
+
+func (e *prometheusExporter) HTTPHandler() http.Handler {
+	return prometheus.HTTPHandler()
+}
+
+func newConstLabels() map[string]string {
```
louyuting, line 108: is it better to define as a const map?

exporter/metric/exporter.go
```diff
@@ -0,0 +1,146 @@
...
+	// NewGauge creates a Gauge metric partitioned by the given label names.
+	NewGauge(name, desc string, labelNames []string) Gauge
+	// NewHistogram creates a histogram metric partitioned by the given label names.
+	NewHistogram(name, desc string, buckets []float64, labelNames []string) Histogram
+	// HTTPHandler returns http.Handler used to export metrics.
+	HTTPHandler() http.Handler
+}
+
+// prometheusExporter is the exporter of prometheus implementation.
+type prometheusExporter struct{}
```
louyuting, line 86: Maybe we could move prometheusExporter related function to pkg `exporter/metric/prometheus`?

core/stat/stat_slot.go
```diff
@@ -60,6 +95,12 @@ func (s *Slot) OnCompleted(ctx *base.EntryContext) {
 	if ctx.Resource.FlowType() == base.Inbound {
 		s.recordCompleteFor(InboundNode(), ctx.Input.BatchCount, rt, ctx.Err())
 	}
+
+	completeCounter.Add(float64(ctx.Input.BatchCount), ctx.Resource.Name())
+	if ctx.Err() != nil {
```
sanxun0325, line 100: When ctx.isBlocked()=true and ctx.err ()!= nil, errorCounter can't count

core/circuitbreaker/circuit_breaker.go
```diff
@@ -194,6 +210,8 @@ func (b *circuitBreakerBase) fromHalfOpenToOpen(snapshot interface{}) bool {
 		for _, listener := range stateChangeListeners {
 			listener.OnTransformToOpen(HalfOpen, *b.rule, snapshot)
 		}
+
+		stateChangeCounter.Add(float64(1), b.BoundRule().Resource, "HalfOpen", "Open")
```
liqiangz, line 214: Maybe we should call stateChangeCounter.Add() in fromHalfOpenToClosed method.

exporter/metric/exporter.go
```diff
@@ -0,0 +1,146 @@
...
+	host      string
+	app       string
+	pid       string
+	namespace string
+)
+
+func init() {
+	exporter = newPrometheusExporter()
+
+	host, _ = os.Hostname()
```
liqiangz, line 38: Maybe we can print the error here.

core/stat/stat_slot.go
```diff
@@ -26,8 +27,38 @@ const (
...
+		[]string{"resource"})
+	rtHistogram = metric_exporter.NewHistogram(
+		"rt",
+		"Rt histogram",
+		[]float64{1.0, 5.0, 10.0, 50.0, 100.0, 500.0, 1000, 5000},
+		[]string{"resource"})
 )
 
+func init() {
+	metric_exporter.MustRegister(passCounter)
```
sczyh30, line 55: Here the process may panic if an error occurred during metric registering. Actually the metric exporter is not mandatory, so maybe we could provide a better mechanism to enable the exporter and handling errors?

iDube:
any progress on this feature？

exporter/metric/exporter.go
luckyxiaoqiang: May be supported in future.

exporter/metric/exporter.go
luckyxiaoqiang: Updated.

exporter/metric/exporter.go
luckyxiaoqiang: It may cause import cycle.

exporter/metric/exporter.go
luckyxiaoqiang: It can be ignored here. We do not care it.

core/config/entity.go
```diff
@@ -46,6 +48,19 @@ type SentinelConfig struct {
...
 
+// ExporterConfig represents configuration items related to exporter, like metric exporter.
+type ExporterConfig struct {
+	Metric MetricExporterConfig
+}
+
+// MetricExporterConfig represents configuration of metric exporter.
+type MetricExporterConfig struct {
+	// HTTPAddr is the http server listen address, like ":8080".
+	HTTPAddr string `yaml:"http_addr"`
```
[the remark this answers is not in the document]
luckyxiaoqiang: Updated.

core/stat/stat_slot.go
luckyxiaoqiang: Currently removed.

core/stat/stat_slot.go
luckyxiaoqiang: We can ignore the register result for now.

core/circuitbreaker/circuit_breaker.go
luckyxiaoqiang: Good catch. Addressed.

exporter/metric/exporter.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// prometheusExporter is the exporter of prometheus implementation.
+type prometheusExporter struct{}
+
+func newPrometheusExporter() *prometheusExporter {
+	return &prometheusExporter{}
+}
+
+func (e *prometheusExporter) NewCounter(name, desc string, labelNames []string) Counter {
```
sanxun0325, line 95: In Prometheus/maine. go, directly implement NewCounter, NewGauge is that ok?

exporter/metric/exporter.go
sanxun0325: > Maybe we could move prometheusExporter related function to pkg `exporter/metric/prometheus`?

I think so too, `import cycle` can you try to solve?

luckyxiaoqiang:
Ping @louyuting @sczyh30

louyuting (approved):
LGTM

sczyh30 (approved):
LGTM

sanxun0325 (approved):
LGTM

sczyh30:

PR #384, opened 2021-01-20
Embeded independent module in repo for datasource

louyuting:
### Special notes for reviews


louyuting:
@sczyh30  Could you please take a look?

sczyh30:
> @sczyh30 Could you please take a look?

I'll verify it this weekend.

sanxun0325 (approved):
LGTM

.github/workflows/go.yml
```diff
@@ -55,6 +55,14 @@ jobs:
         diff -u <(echo -n) <(gofmt -d -s .)
         diff -u <(echo -n) <(goimports -d .)
         go test -v -race ./... -coverprofile=coverage.txt -covermode=atomic
+        cd ./pkg/datasource/consul
```
sczyh30, line 58: Maybe we could improve here later.

sczyh30 (approved):

sczyh30:
This could bring breaking change (because the pkg path has been changed). We might need to find a good way to tackle the compatibility problem later.

.github/workflows/go.yml
louyuting: echo

louyuting:
> This could bring breaking change (because the pkg path has been changed). We might need to find a good way to tackle the compatibility problem later.

Maybe we could keep release version compatibility. 
These datasource and adapter need release synchronously

PR #388, opened 2021-02-01
feat:block_err add opts func ,add RegistryBlockType func;

cafra:
chore: unit test & bench test

feat:
Add option func for block error to facilitate expansion, and add New and Reset methods;
Add custom registration BlockType function;

refactor:
Optimize performance of blockType String function. detail info in bench function.

### Special notes for reviews

### DIFF
--- a/core/base/block_error.go
+++ b/core/base/block_error.go
@@ -27,6 +27,50 @@ type BlockError struct {
 	snapshotValue interface{}
 }
 
+type BlockErrorOption func(*BlockError)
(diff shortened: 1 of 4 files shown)


cafra:
block err 增加options 模式，方便自定义 slot 使用:
1. 注册错误类型；
2. 提高错误的扩展性；
3. 优化错误的String方法的性能；

sczyh30 (approved):

PR #407, opened 2021-05-20
fix issue #403: reset() func

### LINKED ISSUE #403: what does the reset() of MetricBucket do? seems no references
https://github.com/alibaba/sentinel-golang/blob/c362aaeafd1e500c9812d4961bcea81b25008b45/core/stat/base/metric_bucket.go#L68

another question:
the minrt updating function is not thread safe.
```go
func (mb *MetricBucket) AddRt(rt int64) {
	mb.addCount(base.MetricEventRt, rt)
	if rt < atomic.LoadInt64(&mb.minRt) { // here is not thread safe.
		// Might not be accurate here.
		atomic.StoreInt64(&mb.minRt, rt)
	}
}
```

Howie59:
Fixes #403

### DIFF
--- a/core/stat/base/bucket_leap_array.go
+++ b/core/stat/base/bucket_leap_array.go
@@ -37,6 +37,8 @@ func (bla *BucketLeapArray) NewEmptyBucket() interface{} {
 }
 
 func (bla *BucketLeapArray) ResetBucketTo(bw *BucketWrap, startTime uint64) *BucketWrap {
+	b := bla.currentBucketWithTime(util.CurrentTimeMillis())
+	b.reset()
 	atomic.StoreUint64(&bw.BucketStart, startTime)
 	bw.Value.Store(NewMetricBucket())
 	return bw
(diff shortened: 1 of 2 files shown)


core/stat/base/bucket_leap_array.go
```diff
@@ -37,6 +37,8 @@ func (bla *BucketLeapArray) NewEmptyBucket() interface{} {
 }
 
 func (bla *BucketLeapArray) ResetBucketTo(bw *BucketWrap, startTime uint64) *BucketWrap {
+	b := bla.currentBucketWithTime(util.CurrentTimeMillis())
+	b.reset()
```
luckyxiaoqiang, line 41: What's the purpose of the added line? Please see the below line`bw.Value.Store(NewMetricBucket())`.

Howie59: for reset the old bucket and create new one

luckyxiaoqiang: In my view, `ResetBucketTo ` only intend to reset `bw`.

sczyh30:
Any update on this PR?

PR #415, opened 2021-07-10
add cpu ,memory calculate for container

HobbyBear:
let sentinel support  container environment.

### DIFF
--- a/core/system_metric/sys_metric_stat.go
+++ b/core/system_metric/sys_metric_stat.go
@@ -16,8 +16,6 @@ package system_metric
 
 import (
 	"bufio"
-	"github.com/pkg/errors"
-	"github.com/shirou/gopsutil/v3/cpu"
 	"log"
 	"os"
 	"strconv"
@@ -25,6 +23,9 @@ import (
 	"sync"
 	"sync/atomic"
 	"time"
+
+	"github.com/pkg/errors"
+	"github.com/shirou/gopsutil/v3/cpu"
 
 	"github.com/alibaba/sentinel-golang/logging"
 	"github.com/alibaba/sentinel-golang/metrics"
--- a/core/system_metric/sys_metric_stat.go
+++ b/core/system_metric/sys_metric_stat.go
@@ -165,7 +165,7 @@ func retrieveAndUpdateCpuStat() {
 	currentCpuUsage.Store(cpuPercent)
 }
 
-// getProcessCpuStat gets current process's memory usage in Bytes
+// getProcessCpuStat gets current process's cpu usage in Bytes
 func getProcessCpuStat() (float64, error) {
 	curProcess := currentProcess.Load()
 	if curProcess == nil {
--- a/core/system_metric/sys_metric_stat.go
+++ b/core/system_metric/sys_metric_stat.go
@@ -16,13 +16,12 @@ package system_metric
 
 import (
 	"bufio"
+	"fmt"
 	"os"
 	"strconv"
 	"strings"
 	"sync"
 	"time"
-
-	"github.com/shirou/gopsutil/v3/cpu"
 
 	"github.com/alibaba/sentinel-golang/logging"
 	"github.com/alibaba/sentinel-golang/metrics"
@@ -90,11 +89,8 @@ func init() {
 			logging.Error(err, "Fail to getSysCpuUsage when initializing system metric")
 			return
 		}
-		currentContainerCpuTotal, err = getContainerCpuUsage()
-		if err != nil {
-			logging.Error(err, "Fail to getContainerCpuUsage when initializing system metric")
-			return
-		}
+		currentContainerCpuTotal = getContainerCpuUsage()
+
 		preContainerCpuUsage.Store(currentContainerCpuTotal)
 		preSysTotalCpuUsage.Store(currentSysCpuTotal)
 		onlineContainerCpuCount = getContainerCpuCount()
@@ -286,10 +282,7 @@ func GetContainerCpuStat() (float64, error) {
 	if err != nil {
 		return 0, err
 	}
-	currentContainerCpuTotal, err = getContainerCpuUsage()
-	if err != nil {
-		return 0, err
-	}
+	currentContainerCpuTotal = getContainerCpuUsage()
 
 	preSysTotalCpu := preSysTotalCpuUsage.Load()
 
@@ -305,37 +298,40 @@ func GetContainerCpuStat() (float64, error) {
 }
 
 func getSysCpuUsage() (float64, error) {
-	var (
-		currentSysCpuTotal float64
-	)
-	currentCpuStatArr, err := cpu.Times(false)
-	if err != nil {
-		return 0, err
+	path := "/proc/stat"
+	lines := readLineFromFile(path)
+	for _, line := range lines {
+		parts := strings.Fields(line)
+		switch parts[0] {
+		case "cpu":
+			if len(parts) < 8 {
+				return 0, fmt.Errorf("invalid number of cpu fields")
+			}
+			var totalClockTicks float64
+			for _, i := range parts[1:8] {
+				v, err := strconv.ParseFloat(i, 64)
+				if err != nil {
+					return 0, fmt.Errorf("Unable to convert value %s to int: %s", i, err)
+				}
+				totalClockTicks += v
+			}
+			return totalClockTicks / 100, nil
+		}
 	}
-	for _, stat := range currentCpuStatArr {
-		currentSysCpuTotal = stat.User + stat.System + stat.Idle + stat.Nice + stat.Iowait + stat.Irq +
-			stat.Softirq + stat.Steal + stat.Guest + stat.GuestNice
-	}
-	return currentSysCpuTotal, nil
+	return 0, nil
 }
 
-func getContainerCpuUsage() (float64, error) {
+func getContainerCpuUsage() float64 {
 	path := "/sys/fs/cgroup/cpuacct/cpuacct.usage"
-	f, err := os.Open(path)
-	if err != nil {
-		return 0, err
+	usage := readLineFromFile(path)
+	if len(usage) == 0 {
+		return 0
 	}
-	defer f.Close()
-	reader := bufio.NewReader(f)
-	usage, _, err := reader.ReadLine()
+	ns, err := strconv.ParseFloat(strings.TrimSpace(usage[0]), 64)
 	if err != nil {
-		return 0, err
+		return 0
 	}
-	ns, err := strconv.ParseFloat(strings.TrimSpace(string(usage)), 64)
-	if err != nil {
-		return 0, err
-	}
-	return ns / 1e9, nil
+	return ns / 1e9
 }
 
 // getProcessCpuStat gets current process's cpu usage in Bytes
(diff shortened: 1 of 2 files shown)


HobbyBear:
sentinel 目前计算cpu，内存不支持容器环境，容器环境读取此类信息的文件路径不同，所以我在计算相关指标的代码中加了判断。如果是容器环境，那么我将从sys/fs/cgroup/cpuacct/cpuacct.usage_percpu读取cpu信息，从sys/fs/cgroup/memory/memory.usage_in_bytes获取内存占用信息，然后存到相关的指标变量里，取代了之前只读取宿主机的cpu，内存信息。

luckyxiaoqiang:
Please fix CI.

core/system_metric/sys_metric_stat.go
```diff
@@ -32,6 +38,9 @@ const (
 	NotRetrievedLoadValue     float64 = -1.0
 	NotRetrievedCpuUsageValue float64 = -1.0
 	NotRetrievedMemoryValue   int64   = -1
+	CGroupPath                        = "/proc/1/cgroup"
+	DockerPath                        = "/docker"
```
sczyh30, line 42: What if the container runtime is not Docker? Kubernetes is deprecating Docker as a container runtime after v1.20. Refer https://kubernetes.io/blog/2020/12/02/dont-panic-kubernetes-and-docker/

core/system_metric/sys_metric_stat.go
```diff
@@ -15,7 +15,13 @@
 package system_metric
 
 import (
+	"bufio"
+	"github.com/pkg/errors"
+	"github.com/shirou/gopsutil/v3/cpu"
```
sczyh30, line 20: Please reformat and optimize imports of your code, according to the CI failure.

sczyh30 (changes requested):

sanxun0325 (commented):
It might be better to use gopsutil API directly to get the container data.
reference：https://github.com/shirou/gopsutil/blob/master/v3/docker/docker_linux.go

core/system_metric/sys_metric_stat.go
HobbyBear: image
the next line is the path in Kubernetes.

HobbyBear:
> It might be better to use gopsutil API directly to get the container data.
> reference：https://github.com/shirou/gopsutil/blob/master/v3/docker/docker_linux.go

At first, I wanted to use gopsutil too. but i found that gopsutil need to pass container id to get the container metrics. And The get metric path is relative to the host .
image
when we run our code in container, we may hope that the server running like in host. So the path getting the metrics should be the path in container ,not in the host.

sanxun0325:
> > It might be better to use gopsutil API directly to get the container data.
> > reference：https://github.com/shirou/gopsutil/blob/master/v3/docker/docker_linux.go
> 
> At first, I wanted to use gopsutil too. but i found that gopsutil need to pass container id to get the container metrics. And The get metric path is relative to the host .
> image
> when we run our code in container, we may hope that the server running like in host. So the path getting the metrics should be the path in container ,not in the host.

pass container id to get the container metrics,`base `parameter can be passed in. 
https://github.com/shirou/gopsutil/blob/42d667edb6ce16bc36502a3c5b6d0eb7ac8396ec/v3/docker/docker_linux.go#L140

HobbyBear:
> > > It might be better to use gopsutil API directly to get the container data.
> > > reference：https://github.com/shirou/gopsutil/blob/master/v3/docker/docker_linux.go
> > 
> > 
> > At first, I wanted to use gopsutil too. but i found that gopsutil need to pass container id to get the container metrics. And The get metric path is relative to the host .
> > image
> > when we run our code in container, we may hope that the server running like in host. So the path getting the metrics should be the path in container ,not in the host.
> 
> pass container id to get the container metrics,`base `parameter can be passed in. such as `/sys/fs/cgroup `
> https://github.com/shirou/gopsutil/blob/42d667edb6ce16bc36502a3c5b6d0eb7ac8396ec/v3/docker/docker_linux.go#L140

for example ,if i use gopsutil in sentinel and the sentinel is used in a container .The  path to get metrics  is constructed with container_id and base path like this ,/sys/fs/cgroup/cpu/docker/019edf83f18539db17a432fbc39e3a9448435114d7c62920e21373949bc63f34 ,but the path is  relative to the host ,not in a container .if in a container,the path is not found,and our code is running in a container.
the metric file path  of  host is 
image

the metric file path of container is 
image

HobbyBear:
> Please fix CI.

yeap,i fixed it

sanxun0325:
> > It might be better to use gopsutil API directly to get the container data.
> > reference：https://github.com/shirou/gopsutil/blob/master/v3/docker/docker_linux.go
> 
> At first, I wanted to use gopsutil too. but i found that gopsutil need to pass container id to get the container metrics. And The get metric path is relative to the host .
> image
> when we run our code in container, we may hope that the server running like in host. So the path getting the metrics should be the path in container ,not in the host.

The CPU metrics are obtained by passing in the corresponding container ID in the container environment.
`minikube ` is my container.
image

core/system_metric/sys_metric_stat.go
```diff
@@ -156,16 +263,95 @@ func InitCpuCollector(intervalMs uint32) {
...
+	}
+	preContainerCpu, ok := preContainerCpuUsage.Load().(float64)
+
+	if !ok {
+		return 0, errors.New("preContainerCpuUsage load is not float64")
+	}
+	if currentSysCpuTotal-preSysTotalCpu == 0 {
+		return 0, err
+	}
+	return (currentContainerCpuTotal - preContainerCpu) * onlineContainerCpuCount / (currentSysCpuTotal - preSysTotalCpu), err
```
ZoeShaw101, line 317: 没有看到再去更新preContainerCpu preSysTotalCpu了？

HobbyBear: retrieveAndUpdateCpuStat 方法里我加了更新逻辑，retrieveAndUpdateCpuStat是在定时器里执行

HobbyBear:
```
func InitCpuCollector(intervalMs uint32) {
	if intervalMs == 0 {
		return
	}
	cpuStatCollectorOnce.Do(func() {
		// Initial memory retrieval.
		retrieveAndUpdateCpuStat()

		ticker := util.NewTicker(time.Duration(intervalMs) * time.Millisecond)
		go util.RunWithRecover(func() {
			for {
				select {
				case <-ticker.C():
					retrieveAndUpdateCpuStat()
				case <-ssStopChan:
					ticker.Stop()
					return
				}
			}
		})
	})
}
```

core/system_metric/sys_metric_stat.go
```diff
@@ -156,16 +263,95 @@ func InitCpuCollector(intervalMs uint32) {
...
+	currentSysCpuTotal, err = getSysCpuUsage()
+	if err != nil {
+		return 0, err
+	}
+	currentContainerCpuTotal, err = getContainerCpuUsage()
+	if err != nil {
+		return 0, err
+	}
+
+	preSysTotalCpu, ok := preSysTotalCpuUsage.Load().(float64)
```
sanxun0325, line 305: when preSysTotalCpuUsage is nil,panic occurs when float64 is converted

core/system_metric/sys_metric_stat.go
```diff
@@ -156,16 +263,95 @@ func InitCpuCollector(intervalMs uint32) {
...
+	if !ok {
+		return 0, errors.New("preSysTotalCpuUsage load is not float64")
+	}
+	preContainerCpu, ok := preContainerCpuUsage.Load().(float64)
```
sanxun0325, line 309: when preContainerCpuUsage is nil,panic occurs when float64 is converted

core/system_metric/sys_metric_stat.go
```diff
@@ -49,6 +58,11 @@ var (
 	TotalMemorySize    = getTotalMemorySize()
 
 	ssStopChan = make(chan struct{})
+
+	isContainer             bool
+	preSysTotalCpuUsage     atomic.Value
+	preContainerCpuUsage    atomic.Value
```
luckyxiaoqiang, line 64: How about `atomic.Float64`?("go.uber.org/atomic")

core/system_metric/sys_metric_stat.go
```diff
@@ -156,16 +263,95 @@ func InitCpuCollector(intervalMs uint32) {
...
+	f, err := os.Open(path)
+	if err != nil {
+		return 0, err
+	}
+	defer f.Close()
+	reader := bufio.NewReader(f)
+	usage, _, err := reader.ReadLine()
+	if err != nil {
+		return 0, err
+	}
```
luckyxiaoqiang, line 346: Repeated codes. Maybe we can extract a method like `readLineFromFile`.

core/system_metric/sys_metric_stat.go
```diff
@@ -156,16 +263,95 @@ func InitCpuCollector(intervalMs uint32) {
...
+	if !ok {
+		return 0, errors.New("preSysTotalCpuUsage load is not float64")
+	}
+	preContainerCpu, ok := preContainerCpuUsage.Load().(float64)
+
+	if !ok {
+		return 0, errors.New("preContainerCpuUsage load is not float64")
+	}
+	if currentSysCpuTotal-preSysTotalCpu == 0 {
+		return 0, err
```
luckyxiaoqiang, line 315: `return 0, nil`?

core/system_metric/sys_metric_stat.go
luckyxiaoqiang: `return (currentContainerCpuTotal - preContainerCpu) * onlineContainerCpuCount / (currentSysCpuTotal - preSysTotalCpu), nil`?

core/system_metric/sys_metric_stat.go
HobbyBear: the method of calculating cpu is similar to docker stats ,refer this 
`https://stackoverflow.com/questions/45196440/cpupercent-metric-from-docker-stats-vs-cgroups`

core/system_metric/sys_metric_stat.go
```diff
@@ -64,6 +78,68 @@ func init() {
...
+	}
+	defer f.Close()
+	reader := bufio.NewReader(f)
+	usage, _, err := reader.ReadLine()
+	if err != nil {
+		return 0, err
+	}
+	perCpuUsages := strings.Fields(string(usage))
+
+	return float64(len(perCpuUsages)), nil
```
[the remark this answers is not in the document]
HobbyBear: no,it is the cpu count of the container

core/system_metric/sys_metric_stat.go
```diff
@@ -64,6 +78,68 @@ func init() {
...
+	buff := bufio.NewReader(f)
+	for {
+		line, _, err := buff.ReadLine()
+		if err != nil {
+			return false
+		}
+		if strings.Contains(string(line), DockerPath) ||
+			strings.Contains(string(line), KubepodsPath) {
+			return true
+		}
```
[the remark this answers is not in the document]
HobbyBear: yes,  it is running in docker container or in k8s pods ,if in lxc ,maybe we can add new path /lxc ,but  i have not really run in lxc, i need to practice with that.

core/system_metric/sys_metric_stat.go
HobbyBear: yeap,fixed it

HobbyBear:
The cpu utilization rate obtained in sentinel is the same as the mechanism displayed by top command, and 4 cores show 400%, so the cpu utilization rate of the process obtained inside the container is still valid, so I closed this merge.

PR #420, opened 2021-07-26
Add sampleCount zero checking in NewLeapArray func

tylitianrui:
first param of function ` NewLeapArray`   can  not  be  zero.  Or  it  will  cause panic  `runtime error: divide by zero`

### DIFF
--- a/core/stat/base/leap_array.go
+++ b/core/stat/base/leap_array.go
@@ -159,7 +159,7 @@ type LeapArray struct {
 }
 
 func NewLeapArray(sampleCount uint32, intervalInMs uint32, generator BucketGenerator) (*LeapArray, error) {
-	if intervalInMs%sampleCount != 0 {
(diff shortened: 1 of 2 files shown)


core/stat/base/leap_array.go
```diff
@@ -159,7 +159,7 @@ type LeapArray struct {
 }
 
 func NewLeapArray(sampleCount uint32, intervalInMs uint32, generator BucketGenerator) (*LeapArray, error) {
-	if intervalInMs%sampleCount != 0 {
+	if sampleCount == 0 || intervalInMs%sampleCount != 0 {
```
sanxun0325, line 162: Checks that are not zero are checked in the calling function `getRuleStatSlidingWindowBucketCount(r *Rule)` .
Is it necessary to check again?

sczyh30 (approved):
LGTM

sczyh30:
Thanks for contributing!

PR #429, opened 2021-09-15
atomicBucketWrapArray supports 32-bit os

sanxun0325:
On 32-bit os, pointers occupy 4 bytes, currently 8 bytes by default, resulting in some empty buckets

### Special notes for reviews


sczyh30 (approved):
LGTM

sczyh30:
Thanks!

PR #432, opened 2021-10-13
feat: remove sync.Once, avoiding mutex

tylitianrui:
`sync.Once`  is only really needed if what you are doing inside it is really required for the rest of the code to continue. This is not the case here.

### DIFF
--- a/core/flow/slot.go
+++ b/core/flow/slot.go
@@ -15,6 +15,8 @@
 package flow
 
 import (
+	"sync/atomic"
+
 	"github.com/alibaba/sentinel-golang/core/base"
 	"github.com/alibaba/sentinel-golang/core/stat"
 	"github.com/alibaba/sentinel-golang/logging"
@@ -85,9 +87,9 @@ func selectNodeByRelStrategy(rule *Rule, node base.StatNode) base.StatNode {
 func checkInLocal(tc *TrafficShapingController, resStat base.StatNode, batchCount uint32, flag int32) *base.TokenResult {
 	actual := selectNodeByRelStrategy(tc.rule, resStat)
 	if actual == nil {
-		logging.FrequentErrorOnce.Do(func() {
+		if ok := atomic.CompareAndSwapUint32(&logging.FrequentErrorOnce, 0, 1); ok {
 			logging.Error(errors.Errorf("nil resource node"), "No resource node for flow rule in FlowSlot.checkInLocal()", "rule", tc.rule)
-		})
+		}
 		return base.NewTokenResultPass()
 	}
 	return tc.PerformChecking(actual, batchCount, flag)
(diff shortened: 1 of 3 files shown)


tylitianrui:
> Could you please fix the CI failure?

i  have  no  idea how  to fix it .   the  type of  ` FrequentErrorOnce`  has been  changed , from `sync.Once`  to  `uint32`, in  my PR .  IT  is  not  merged  yet.    ` FrequentErrorOnce`  is still   a `sync.Once`   in  `alibaba:master`


`github.com/alibaba/sentinel-golang/logging`   is  required   by  `module github.com/alibaba/sentinel-golang/pkg/datasource/consul`.  logging.FrequentErrorOnce ,  in  repo  `https://github.com/alibaba/sentinel-golang`,   is still  a   `sync.Once`   before  merging [PR432](https://github.com/alibaba/sentinel-golang/pull/432).  so "type error" is  reported  in testing

pkg/datasource/consul/consul.go
```diff
@@ -107,9 +108,9 @@ func (c *consulDataSource) watch() {
 				time.Sleep(time.Second)
 				continue
 			}
-			logging.FrequentErrorOnce.Do(func() {
+			if ok := atomic.CompareAndSwapUint32(&logging.FrequentErrorOnce, 0, 1); ok {
```
sanxun0325, line 111: logging.FrequentErrorOnce used in multiple places, one place cas changes to 1, the others are never executed

core/flow/slot.go
```diff
@@ -85,9 +87,9 @@ func selectNodeByRelStrategy(rule *Rule, node base.StatNode) base.StatNode {
 func checkInLocal(tc *TrafficShapingController, resStat base.StatNode, batchCount uint32, flag int32) *base.TokenResult {
 	actual := selectNodeByRelStrategy(tc.rule, resStat)
 	if actual == nil {
-		logging.FrequentErrorOnce.Do(func() {
+		if ok := atomic.CompareAndSwapUint32(&logging.FrequentErrorOnce, 0, 1); ok {
```
sanxun0325, line 90: What is the difference between this modification and sync.once？

pkg/datasource/consul/consul.go
tylitianrui: yes，same as  `logging.FrequentErrorOnce.Do` .  once ` logging.FrequentErrorOnce.Do` was executed , the others are never executed

core/flow/slot.go
tylitianrui: `sync.Once` 
 this code block  are executed in two or more goroutines, only one can  execute , the others  must wait   it  finished  
 ,then, continue.  `sync.Once`    blocks the rest 

cas  is non-block

PR #448, opened 2021-12-25
fix getProcessCpuStat  uncompatibility

### LINKED ISSUE #435: [BUG] getProcessCpuStat return value (in percentage) breaks compatibility
## Issue Description

Type: *bug report*

### Describe what happened

getProcessCpuStat return value (in percentage) breaks compatibility. In previous implementation, the CPU usage representation is the value between [0.0, 1.0]. Since 1.0.x, the CPU usage from getProcessCpuStat was changed to percentage value (i.e. 0-100), which breaks compatibility and affects system adaptive rules.

### Describe what you expected to happen

`getProcessCpuStat` keeps compatibility and unifys the CPU usage value to [0.0, 1.0]

### Tell us your environment

Sentinel Go v1.0.3

xiaoshi100086:
[[BUG] getProcessCpuStat return value (in percentage) breaks compatibility](https://github.com/alibaba/sentinel-golang/issues/435)

Fixes #435 

### Describe how you did it
I change variables "currentCpuUsage" value which representing CPU usage percentage as value which representing CPU usage as [0, 1.0]. This code fix "getProcessCpuStat" uncompatibility bug.

### Special notes for reviews
"getProcessCpuStat" return multiple  CPU cores usage percentage. That is to say, the value can also be > 100.0 in case of a process running multiple threads on different CPUs. 

So I normalizate this value to [0, 1.0] for the following reasons:
1. ease of use. The user should not be related to the cpu number of the machine.
2. compatibility to System Adaptive Protection threshold. This threshold num should be in [0, 1.0] when rule type is CPUusage，


core/system_metric/sys_metric_stat.go
```diff
@@ -178,7 +180,11 @@ func retrieveAndUpdateCpuStat() {
 
 	cpuRatioGauge.Set(cpuPercent)
 
-	currentCpuUsage.Store(cpuPercent)
+	numcpu := runtime.NumCPU()
```
sanxun0325, line 183: why divide by the cpu num？

xiaoshi100086: "getProcessCpuStat" return multiple CPU cores usage percentage. That is to say, the value can also be > 100.0 in case of a process running multiple threads on different CPUs.

It divide by the cpu num to normalizate this value to [0, 1.0]. For the following reasons:

1. ease of use. The user should not be related to the cpu number of the machine.
2. compatibility to System Adaptive Protection threshold value. This threshold value should be in [0, 1.0] when rule type is CPUusage.

core/system_metric/sys_metric_stat.go
```diff
@@ -178,7 +180,11 @@ func retrieveAndUpdateCpuStat() {
...
+	cpuUsage := cpuPercent / float64(numcpu) / float64(100)
+	cpuUsage = math.Max(0.0, cpuUsage)
```
sanxun0325, line 185: When can it be less than 0.0 or greater than 1.0

xiaoshi100086: I find process.Percent in gopsutil library may return very large invalid value when have mary goroutines, this will cause cpuUsage > 1. So I add code "cpuUsage = math.Min(1.0, cpuUsage)". 
I submit issure in gopsutil to report bug, https://github.com/shirou/gopsutil/issues/1214.

process.Percent can not return value <0, the code "cpuUsage = math.Max(0.0, cpuUsage)" is useless and redundant.

sanxun0325: cpuUsage = math.Max(0.0, cpuUsage) is not required， can be removed

sanxun0325 (commented):
Please sign the CLA

xiaoshi100086:
> 

done

sanxun0325 (approved):

bbiao:
@louyuting @sczyh30 I encountered the same problem fixed in this PR, and I think this bug is important to fix as it affects the functionality of the code. Would you please review this and release a new version?

Issue #451, opened 2022-01-04
[Integration] Add adapter for go-zero


sczyh30:
## Issue Description

Type: *feature request*

### Describe what feature you want

Add adapter for [go-zero](https://github.com/tal-tech/go-zero).

PR #458, opened 2022-02-08
Add adapter for go-zero

Forsworns:

#451

#451

### Describe how you did it

The implementation is based on go-zero middlewares for http servers. 

- Provide an adapter for go-zero global middleware.
- Provide a go-zero routing middleware example.

They have been tested in `middleware_test.go`.

go-zero zrpc simply wraps grpc, and can reuse the interceptors for grpc

### Describe how to verify it

The implementation can be verified via `go test`.  ~~The testcases have to be run one by one manually in case of contention on port.~~

Takeaways:
- The go-zero server did not implement  `http.Handler`. It cannot be tested via `httptest`. 
- And it cannot be stopped in codes. (The `go-zero/rest.Server.Stop` only closes its logger.)

### Special notes for reviews


sczyh30 (approved):
LGTM

sczyh30:

Issue #450, opened 2022-01-04
[Integration] Add adapter for Kratos


sczyh30:
## Issue Description

Type: *feature request*

### Describe what feature you want

Add adapter for https://github.com/go-kratos/kratos

Forsworns:
I'll take this in these days, too. Have found examples on [Kratos middleware](https://github.com/go-kratos/kratos/blob/main/examples/middleware/main.go).

PR #459, opened 2022-02-11
Kratos

Forsworns:

#450

#450

### Describe how you did it

The implementation is based on Kratos unified middleware abstractions for http/grpc servers/clients.

Tests on HTTP servers/clients are provided in `server_test.go` and `client_test.go` .

### Describe how to verify it

The implementation can be verified via `go test`.

### Special notes for reviews


Casper-Mars:
Why fallback only return error without response?

Forsworns:
> Why fallback only return error without response?

- I found all other adapters in Sentinel only return an `error` in their fallback options. I don't know if a response return value would be useful. Of course, it is more flexible :) 
- I found `reply` and `error` are mutually exclusive in some [Kratos examples](https://github1s.com/go-kratos/kratos/blob/HEAD/examples/http/middlewares/services.go). Thus, I assume that the `reply` is `nil` when `blockFallback` is called in Sentinel middleware.
- If necessary, the `kratos/errors` contains status `code`, `reason` and `message` fields. Informative responses can be appended in the `message` field.

Casper-Mars:
> > Why fallback only return error without response?
> 
> * I found all other adapters in Sentinel only return an `error` in their fallback options. I don't know if a response return value would be useful. Of course, it is more flexible :)
> * I found `reply` and `error` are mutually exclusive in some [Kratos examples](https://github1s.com/go-kratos/kratos/blob/HEAD/examples/http/middlewares/services.go). Thus, I assume that the `reply` is `nil` when `blockFallback` is called in Sentinel middleware.
> * If necessary, the `kratos/errors` contains status `code`, `reason` and `message` fields. Informative responses can be appended in the `message` field.

Hi~Thanks for reply. The reason that `reply` is `nil` in kratos middleware is because those middleware doesn`t invoke business logic. So they only need to report error. 
But in practice, the fallback would invoke some business logic. Such as return previous value. 
And I don`t think return a reply within an error is a good idea. This require client to process special error with same business logic. At the same time, it couple results and errors. It make me confused that some other adapters only return an `error`.

sanxun0325 (commented):
LGTM

sanxun0325 (approved):

shenqidebaozi:
You can consider submitting directly to Kratos contrib.

Forsworns:
> You can consider submitting directly to Kratos contrib.

👍  I'll test it on the latest Kratos in these days and try to submit there

oss-sentinel-ai:
⚠️ **Merge conflict detected**

This PR has conflicts with the `master` branch and cannot be merged. Please rebase or merge `master` into your branch and resolve the conflicts:

```bash
git fetch origin
git checkout kratos
git rebase origin/master
# resolve conflicts, then:
git push --force-with-lease
```

This is a one-time reminder. Feel free to @mention me for a re-review after conflicts are resolved.

---
*Automated notification by github-manager-bot*

PR #523, opened 2023-04-12
fix: function comment typo

horoc:
### Describe what this PR does / why we need it

image

### Does this pull request fix one issue?

yes

### Describe how you did it

fix comment

### Describe how to verify it

None

### Special notes for reviews

None

PR #537, opened 2023-06-25
function comment typo(try ci)

binbin0325:
### Special notes for reviews


binbin0325:
replace https://github.com/alibaba/sentinel-golang/pull/523, purpose is to verify CI

PR #541, opened 2023-08-04
add load nacos hotspot param datasource_nacos_hotspot_param_rules_exa…

wenlong139:
…mple.go

### Special notes for reviews


binbin0325 (commented):
It may not be necessary， the hotspot demo is under example pkg

PR #579, opened 2024-09-21
[feature] add outlier ejection

wooyang2018:
关联issue：https://github.com/alibaba/sentinel-golang/issues/575


binbin0325 (commented):
this capability should be a function of the registry center, which can sense if the node is still alive/healthy. It can be discussed further

EndlessSeeker:
1. it is recommended that the outlier slot be optional. By default, the slot will not be executed unless specified to protect the original logic
2. retryer and recycler need to be refactored to be more asynchronous to avoid blocking the core logic

LearningGp (approved):
LGTM

Issue #580, opened 2024-09-24
[BUG] Always an EmptyExporter


binbin0325:
## Issue Description
Read the httpaddr in the exporter init function is always a null character, because the global configuration has not been set at the time of the init function execution https://github.com/alibaba/sentinel-golang/blob/master/exporter/metric/exporter.go#L36

Type: *bug report*

### Describe what happened
empty exporter it doesn't make sense, judgment has been made in the initCoreComponents func

### How to reproduce it (as minimally and precisely as possible)

1. 
2. 
3.

PR #581, opened 2024-09-24
del empty exporter

binbin0325:
empty exporter it doesn't make sense, judgment has been made in the initCoreComponents func

fix:https://github.com/alibaba/sentinel-golang/issues/580

### Special notes for reviews


exporter/metric/exporter.go
```diff
@@ -33,11 +33,7 @@ var (
 )
 
 func init() {
-	if config.MetricExportHTTPAddr() != "" {
-		exporter = newPrometheusExporter()
-	} else {
-		exporter = newEmptyExporter()
-	}
+	exporter = newPrometheusExporter()
```
oss-sentinel-ai, line 36: With this change the prometheus exporter is active even when `MetricExportHTTPAddr` is empty: metrics get registered/collected (though not exposed, since the HTTP server in `api/init.go` only starts when the addr is configured). Could you confirm this is intended? Apps that never export metrics will now pay a small collection overhead. If intended, a short note in the PR description/docs would help.

oss-sentinel-ai (commented):
### Summary
Removes `EmptyExporter` and always initializes the prometheus exporter. Verified there are no other references to `EmptyExporter`/`newEmptyExporter` in the repo, and the metrics HTTP endpoint is still only registered when `MetricExportHTTPAddr` is configured (`api/init.go`), so nothing is exposed unintentionally. One behavioral question inline regarding collection overhead when metric export is disabled.

### Findings
- **[Info]** exporter/metric/exporter.go:36 — confirm intended behavior when export addr is unset (see inline comment)

---

PR #591, opened 2025-01-18
chore: remove unnecessary time convert

qshuai:
remove unnecessary time convert

None

### Describe how you did it
To  get current time in milliseconds, just call time.Now().UnixMilli() directly instead of time.Now().UnixNano() / (time.Millisecond / time.Nanosecond). 

### Describe how to verify it
Clearly

### Special notes for reviews

### DIFF
--- a/util/time.go
+++ b/util/time.go
@@ -21,9 +21,8 @@ import (
 )
 
 const (
-	TimeFormat         = "2006-01-02 15:04:05"
-	DateFormat         = "2006-01-02"
-	UnixTimeUnitOffset = uint64(time.Millisecond / time.Nanosecond)
+	TimeFormat = "2006-01-02 15:04:05"
+	DateFormat = "2006-01-02"
 )
 
 var (
@@ -84,7 +83,7 @@ func (t *RealClock) CurrentTimeMillis() uint64 {
 	if tickerNow > uint64(0) {
 		return tickerNow
 	}
-	return uint64(time.Now().UnixNano()) / UnixTimeUnitOffset
+	return uint64(time.Now().UnixMilli())
 }
 
 func (t *RealClock) CurrentTimeNano() uint64 {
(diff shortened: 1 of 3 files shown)


util/time.go
```diff
@@ -21,9 +21,8 @@ import (
 )
 
 const (
-	TimeFormat         = "2006-01-02 15:04:05"
-	DateFormat         = "2006-01-02"
-	UnixTimeUnitOffset = uint64(time.Millisecond / time.Nanosecond)
+	TimeFormat = "2006-01-02 15:04:05"
+	DateFormat = "2006-01-02"
```
oss-sentinel-ai, line 25: Compatibility concern: this removes the exported constant `UnixTimeUnitOffset`. The `util` package is importable by downstream users (adapters and embedded consumers), so removing an exported identifier is a breaking API change — code referencing `util.UnixTimeUnitOffset` will fail to compile after upgrading. Consider keeping it for at least one release with a `// Deprecated: use time.UnixMilli()/time.Millisecond directly` note, or call this out explicitly in the release notes if the removal is intentional.

util/time.go
```diff
@@ -84,7 +83,7 @@ func (t *RealClock) CurrentTimeMillis() uint64 {
 	if tickerNow > uint64(0) {
 		return tickerNow
 	}
-	return uint64(time.Now().UnixNano()) / UnixTimeUnitOffset
+	return uint64(time.Now().UnixMilli())
```
oss-sentinel-ai, line 86: The substitution itself is correct: `time.Now().UnixMilli()` is semantically identical to `uint64(time.Now().UnixNano()) / 1e6` (Go >= 1.17, go.mod requires 1.18) and avoids the int64 nanosecond overflow horizon issue entirely. `FormatTimeMillis`/`FormatDate` keep identical behavior as well.

oss-sentinel-ai (commented):
### Summary
Replaces the hand-rolled `UnixNano()/1e6` millisecond conversions with `time.Time.UnixMilli()` and removes the `UnixTimeUnitOffset` constant. The conversions are semantically identical (verified including `FormatTimeMillis`/`FormatDate`), and `UnixMilli` additionally sidesteps the int64 nanosecond overflow horizon. One compatibility concern before merging: `UnixTimeUnitOffset` is an exported identifier and removing it breaks any downstream code referencing `util.UnixTimeUnitOffset` — see the inline suggestion (deprecate first, or document the break).

---

PR #592, opened 2025-01-21
refactor(stat): 优化 LeapArray 数据结构并移除未使用的 util 文件

578223592:
- 在 leap_array.go 中，使用更简洁的方式计算 base地址
- 删除整个 util/safe.go 文件，该文件未被使用

### Describe how to verify it
正确性保证可以参考：https://gist.github.com/578223592/3dc108739a91f59e6695fe0368e00d42

### Special notes for reviews


oss-sentinel-ai (commented):
### Summary
Replaces the hand-rolled `SliceHeader` pointer arithmetic in `NewAtomicBucketWrapArrayWithTime` with a direct `unsafe.Pointer(&ret.data[0])`, and removes the now-unused `util.SliceHeader`/`util.StringHeader` types. Verified no remaining references to those types elsewhere in the repo, so the deletion is safe; the new form is semantically equivalent to reading `SliceHeader.Data`.

Note: no CLA status context was reported for this PR head, so holding off on approval until the CLA state is clear — code-wise this looks good.

### Findings
- **[Info]** core/stat/base/leap_array.go:93 — zero-length edge case (see inline comment)

---

core/stat/base/leap_array.go
```diff
@@ -90,9 +90,7 @@ func NewAtomicBucketWrapArrayWithTime(len int, bucketLengthInMs uint32, now uint
 		startTime += uint64(bucketLengthInMs)
 	}
 
-	// calculate base address for real data array
-	sliHeader := (*util.SliceHeader)(unsafe.Pointer(&ret.data))
-	ret.base = unsafe.Pointer((**BucketWrap)(unsafe.Pointer(sliHeader.Data)))
+	ret.base = unsafe.Pointer(&ret.data[0])
```
oss-sentinel-ai, line 93: The simplified `unsafe.Pointer(&ret.data[0])` is semantically equivalent to the old `SliceHeader.Data` computation, nice cleanup. One minor edge case: indexing `data[0]` panics if `len == 0`, whereas the previous header-based form did not index the slice. All known callers pass a positive length, so this is not blocking — a guard would only be needed if zero-length arrays ever become valid.

